PR #4, opened 2020-01-07
Improved the example app UI, its usage and logging

PieterAelse:
Improved the example app: 

- Better UI
   - Actions can only be done when state is correct (eg. device is connected)
- Better logging
- Fixed error in streams

| App started | Scan started | Device detail screen |
| --- | --- | --- |
| flutter_reactive_ble_example_app_1 | flutter_reactive_ble_example_app_2 | flutter_reactive_ble_example_app_3 |


PieterAelse:
@remonh87 I think it's better to only look at the end-code instead of comparing it to what it was.

example/lib/src/utils.dart
```diff
@@ -0,0 +1 @@
+void log(String text) => print("[FlutterReactiveBLE] $text");
```
remonh87: nit: maybe name it FlutterReactiveBLEAPP  to distinguish between library and app>

example/lib/ui/device_detail_screen.dart
```diff
@@ -0,0 +1,159 @@
...
+        deviceId: deviceId,
+      ));
+    });
+
+    _currentValueUpdateSubscription = _charValueStream.listen((data) => _currentValue = data.first);
+
+    _connectedDeviceStream
+        .where((device) => device.connectionState == DeviceConnectionState.connected)
+        .first
+        .then((_) => Future<void>.delayed(const Duration(milliseconds: 20)))
```
remonh87: For Android maybe we should increase the delay to 50 or 100ms personally think 20 is a bit to short

example/lib/ui/device_detail_screen.dart
```diff
@@ -0,0 +1,159 @@
...
+      characteristicId: _charUuid,
+      serviceId: _serviceUuid,
+      deviceId: deviceId,
+    ));
+
+    setState(() => _currentValue = result.first);
+    log("Characteristic read: $_currentValue");
+  }
+
+  Future<void> _writeCharacteristic() async {
```
remonh87: I think we should give users the chance to choose a value to write but for now it can be enough

example/lib/ui/device_list.dart
```diff
@@ -0,0 +1,149 @@
+import 'dart:async';
+
+import 'package:flutter/material.dart';
+import 'package:flutter_reactive_ble/flutter_reactive_ble.dart';
+
+import '../src/utils.dart';
+import '../src/widgets.dart';
+import 'device_detail_screen.dart';
+
+final _uuid = Uuid.parse('0000FE0F-0000-1000-8000-00805F9B34FB');
```
remonh87: I think we should let the user fil in the uuid since this will be only related to lights and this is a general purpose library,

PieterAelse: Should this be the default value when input is empty?

remonh87: No better give an error since we really require to filter on an uuid

PR #11, opened 2020-02-04
add rssi to discovered device

remonh87:


lib/src/model/discovered_device.dart
```diff
@@ -27,7 +27,14 @@ class DiscoveredDevice extends $DiscoveredDevice {
...
 
-  const DiscoveredDevice({@required this.id, @required this.name, @required this.serviceData});
+  @CustomEquality(Ignore())
+  final int rssi;
+
+  const DiscoveredDevice(
+      {@required this.id,
+      @required this.name,
+      @required this.serviceData,
+      @required this.rssi});
```
keke-dandois: add a ,

lib/src/converter/protobuf_converter.dart
```diff
@@ -31,7 +31,10 @@ class ProtobufConverter {
     return ScanResult(
       result: resultFrom(
           getValue: () => DiscoveredDevice(
-              id: message.id, name: message.name, serviceData: serviceData),
+              id: message.id,
+              name: message.name,
+              serviceData: serviceData,
+              rssi: message.rssi),
```
keke-dandois: add a ,

lib/src/model/discovered_device.dart
```diff
@@ -27,7 +27,14 @@ class DiscoveredDevice extends $DiscoveredDevice {
   @CustomEquality(DeepCollectionEquality())
   final Map<Uuid, Uint8List> serviceData;
 
-  const DiscoveredDevice({@required this.id, @required this.name, @required this.serviceData});
+  @CustomEquality(Ignore())
```
miguelcmedeiros: Why is this field ignored for equality?

remonh87: My reasoning behind it is that a device with the same id, service data and name but with a different RSSI would be seen as a different object and I think this shouldn't be the case.

Issue #16, opened 2020-02-20
Use new android embedding 2.0


remonh87:
https://flutter.dev/docs/development/packages-and-plugins/plugin-api-migration

PR #17, opened 2020-02-20
move to new android binding for plugin

remonh87:


android/src/main/kotlin/com/signify/hue/flutterreactiveble/PluginController.kt
```diff
@@ -77,7 +78,7 @@ class PluginController {
         https://github.com/ReactiveX/RxJava/wiki/What's-different-in-2.0#error-handling
         */
         RxJavaPlugins.setErrorHandler { throwable ->
-            if (throwable is UndeliverableException && throwable.cause is BleException) {
+            if (throwable is UndeliverableException && (throwable.cause is BleException) || throwable.cause is RuntimeException) {
```
werediver: I don't think parentheses around `throwable.cause is BleException` are required.

What's the rationale behind adding `throwable.cause is RuntimeException`?

Do you mean what is written? The grouping is as follows: `(throwable is UndeliverableException && throwable.cause is BleException) || throwable.cause is RuntimeException)`.

remonh87: whoops indeed this will provide nasty side effects s oremoved it

remonh87:
Fixes #16

werediver:
As we concluded off-line, it looks like a good idea to _stop_ setting the global RX error handler in our plug-in. Let's track it in a separate issue.

remonh87:
You are right created #18

PR #19, opened 2020-02-20
Improve UUID parsing and validation

werediver:


example/lib/src/ui/device_list.dart
```diff
@@ -56,14 +56,10 @@ class _DeviceListState extends State<DeviceList> {
...
-    if (uuidText.length > 3 && uuidText.length.isEven) {
-      try {
-        Uuid.parse(uuidText);
-        return true;
-      } on Exception {
-        return false;
-      }
-    } else {
+    try {
+      Uuid.parse(uuidText);
```
werediver: Now the validation relies fully on `Uuid.parse(·)` method.

lib/src/model/uuid.dart
```diff
@@ -13,27 +13,31 @@ class Uuid {
...
       if (byte == null) throw _UuidParseFailure(string);
 
       data[byteOffset] = byte;
 
       byteOffset += 1;
       substringStart += 2;
     }
-
-    return Uuid(data.buffer.asUint8List(0, byteOffset));
+    if (byteOffset == 2 || byteOffset == 16) {
```
werediver: Only 2 bytes ([registered](https://www.bluetooth.com/specifications/gatt/services/) short service IDs) and 16 bytes (everything else) (UU)IDs are used in BLE.

werediver: Correction, according to [CBUUID](https://developer.apple.com/documentation/corebluetooth/cbuuid) docs, 2/4/16 bytes are used. Will change this and the unit-tests.

PR #24, opened 2020-02-23
Increase test coverage

remonh87:


remonh87:
Increased test coverage with 10%. Biggest hit will be done once we solved #25

test/protobuf_converter_test.dart
```diff
@@ -183,6 +186,189 @@ void main() {
             result.iif(success: id, failure: (_) => throw Exception()), value);
       });
     });
+
+    group("ConnectionStateUpdate conversion", () {
+      const id = 'id';
+      const connectionState = 1;
+      final message = pb.DeviceInfo()
+        ..id = id
+        ..connectionState = connectionState;
```
werediver: Instances of the classes generated by protobuf are mutable. Ideally, mutable objects should be instantiated in a set-up block to avoid mutual test interference (in case one or more of the tests modify such an object).

test/protobuf_converter_test.dart
```diff
@@ -183,6 +186,189 @@ void main() {
...
+      });
+
+      test('converts failure', () {
+        final failure = pb.GenericFailure()
+          ..code = 0
+          ..message = "failure";
+        final failedUpdate = pb.DeviceInfo()..failure = failure;
+
+        final updateResult =
+            sut.connectionStateUpdateFrom(failedUpdate).failure;
```
werediver: It's confusing that in most of the test cases in this group `sut.connectionStateUpdateFrom(·)` operates on `message`, but here the pattern breaks. I'd suggest separating the test-cases into a few groups so that the test-cases are as uniform in each one as possible.

test/protobuf_converter_test.dart
```diff
@@ -183,6 +186,189 @@ void main() {
...
+        expect(updateResult.message, "failure");
+        expect(updateResult.code, ConnectionError.unknown);
+      });
+
+      test('converts unknown code throws', () {
+        final unknownCodeMessage = message..connectionState = 100;
```
werediver: This is also somewhat confusin. Can find a way to avoid mutating the test conditions initially set up in a group?

test/protobuf_converter_test.dart
```diff
@@ -183,6 +186,189 @@ void main() {
...
+    group('Converts clear gatt cache result', () {
+      test('Succeeds', () {
+        final result = sut.clearGattCacheResultFrom(pb.ClearGattCacheInfo());
+
+        expect(
+            result.iif(
+                success: (_) => "success",
+                failure: (_) => throw AssertionError("Not expected to fail")),
+            "success");
+      });
```
werediver: What about `expect(result, Result.success("success"))`?

test/rx_ext/repeater_test.dart
```diff
@@ -0,0 +1,134 @@
...
+    _RepeaterHandlerMock<int> handler;
+    Repeater<int> sut;
+
+    setUp(() {
+      underlyingStreamController = StreamController(sync: true);
+      handler = _RepeaterHandlerMock();
+
+      when(handler.onListenEmitFrom())
+          .thenAnswer((_) => underlyingStreamController.stream);
+    });
```
werediver: Ouch. The SUT is declared at _this_ level, but initialized. That make the variable useless at _this_ level (if not dangerous, because it is effectively uninitialized [yes, it's "initialized" to `null` but do you expect it to be null in SUT?]).

test/rx_ext/repeater_test.dart
```diff
@@ -0,0 +1,134 @@
...
+
+    group('Repeater internals', () {
+      setUp(() {
+        sut = Repeater(
+          onListenEmitFrom: handler.onListenEmitFrom,
+          onCancel: handler.onCancel,
+          isSync: true,
+        );
+      });
```
werediver: _Usually_ (not saying "always"), it's possible to initialize the SUT _once_ in the top-level group. _Usually_ it takes just a bit more mocking (not sure it's the case here).

test/rx_ext/serial_disposable_test.dart
```diff
@@ -0,0 +1,54 @@
...
+import 'package:mockito/mockito.dart';
+
+abstract class _Disposer {
+  Future<void> dispose(int value);
+}
+
+class _DisposerMock extends Mock implements _Disposer {}
+
+void main() {
+  group('Serialdisposable', () {
```
werediver: Use "$SerialDisposable" instead. More proof to potential refactorings (renames).

test/rx_ext/serial_disposable_test.dart
```diff
@@ -0,0 +1,54 @@
+import 'package:flutter_reactive_ble/flutter_reactive_ble.dart';
+import 'package:flutter_test/flutter_test.dart';
+import 'package:mockito/mockito.dart';
+
+abstract class _Disposer {
+  Future<void> dispose(int value);
+}
+
+class _DisposerMock extends Mock implements _Disposer {}
```
werediver: Mocks are not the goal, tests are the goal. Put what matters higher in the file, so a reader doesn't have to go through the unimportant stuff first.

test/rx_ext/serial_disposable_test.dart
remonh87: I agree that $ makes more sense but it gives problems running tests in VSCode  ( it fails at first after rerun it succeeds )

test/rx_ext/serial_disposable_test.dart
werediver: Are you sure you want to use _that_ ~~IDE~~ editor?.. :D

test/rx_ext/serial_disposable_test.dart
werediver: Are you sure you want to use _that_ ~~IDE~~ editor?.. :D

test/protobuf_converter_test.dart
remonh87: you are right I was lazy but got caught ;)

test/rx_ext/repeater_test.dart
remonh87: I use different constructors for the sut (regular, fromStream and fromBroadcast).  I declared the sut in the respective group so test is more readable now

test/rx_ext/serial_disposable_test.dart
remonh87: I like that aprroach makes sense

test/rx_ext/repeater_test.dart
werediver: That may be considered a wrong abstraction: it saves you some typing, but causes questions, thus not being sound. Yes, I mean being honest and declaring the SUT in multiple groups would be more straightforward and unquestionable. No, I don't insist too much on changing this.

remonh87:
@werediver  can you check again? I fixed the type for clearGattCache but I do not think we need to expose unit to the outside for now right? Can be sth to consider later in 3.0 or sth

test/rx_ext/serial_disposable_test.dart
werediver: I recently learned that you do not have to declare such dumb classes just to mock a function. An alternative is to declare some members right in the mock declaration.

```dart
class Function1Mock<T0, T1> extends Mock {
  T0 call(T0 arg);
}
```

At least we have such declarations in our internal codebase. Maybe `=> null` body can work, didn't try.

lib/src/converter/protobuf_converter.dart
```diff
@@ -1,5 +1,6 @@
 import 'dart:typed_data';
 
+import 'package:flutter_reactive_ble/flutter_reactive_ble.dart';
```
werediver: Nope, don't import the whole library from inside the library.

lib/src/converter/protobuf_converter.dart
```diff
@@ -61,10 +62,10 @@ class ProtobufConverter {
         ),
       );
 
-  Result<void, GenericFailure<ClearGattCacheError>> clearGattCacheResultFrom(
+  Result<Unit, GenericFailure<ClearGattCacheError>> clearGattCacheResultFrom(
```
werediver: Now, this `Unit` type does have some exposure, because it's used in a declaration of a public method. Maybe in Dart it's okay to still not export the definition of this type, but that may cause some lesser or bigger issues on the library user side.

lib/src/model/unit.dart
```diff
@@ -0,0 +1,12 @@
+import 'package:meta/meta.dart';
+
+@immutable
+class Unit {
```
werediver: It's a pity every library needs its own `Unit` (like in [sum_types](https://github.com/werediver/sum_types.dart/blob/19f96753f76c6e9e9f93e8d765d900dded9e6a5e/sum_types/lib/src/support.dart#L4) and now here). But we can rightfully blame the poorly designed Dart SDK 🎉

lib/src/converter/protobuf_converter.dart
remonh87: wow where did that come from?

test/rx_ext/repeater_test.dart
```diff
@@ -0,0 +1,138 @@
...
+      test("unsubscribes from the underlying stream when detached", () async {
+        sut.stream.listen((_) {});
+        await sut.detach();
+
+        verify(handler.onCancel()).called(1);
+        expect(underlyingStreamController.hasListener, false);
+      });
+
+      test("unsubscribes from the underlying stream when disposed", () async {
+        sut.stream.listen((_) {});
```
werediver: This line repeats in multiple test cases. Why not to make it a part of the test conditions by moving it into a set-up block?

lib/src/converter/protobuf_converter.dart
werediver: It's pretty easy to end up with using the suggestions from an IDE.

lib/src/converter/protobuf_converter.dart
remonh87: I previously had it but since the public function is of type void there shouldn't be a problem in Dart. I think it could be a nice idea to change things to Unit later (this will break API ).

werediver:
A general advice on structuring unit-tests: stick to the BDD structure (given [connditions] /when [action] / then [expectations]) as much as possible. In some more rare cases it brings much more overhead then gives benefits, then step away from it with understanding that you're making a use case specific optimization. In most of the cases, though, application of the BDD test structure yields a very readable test suite.

lib/src/converter/protobuf_converter.dart
werediver: Ah, so this one is not exposed? Then sure it's okay to keep `Unit` internal.

test/protobuf_converter_test.dart
```diff
@@ -183,6 +186,204 @@ void main() {
...
+      const id = 'id';
+      const value = [2, 3];
+      final characteristic = pb.CharacteristicAddress()
+        ..deviceId = id
+        ..serviceUuid = (pb.Uuid()..data = [0])
+        ..characteristicUuid = (pb.Uuid()..data = [1]);
+
+      final message = pb.CharacteristicValueInfo()
+        ..characteristic = characteristic
+        ..value = value;
```
werediver: Ideally, non-const stuff should be initialized in a set-up block.

test/protobuf_converter_test.dart
```diff
@@ -183,6 +186,204 @@ void main() {
...
+        final result = sut.characteristicValueFrom(message);
+        expect(result.characteristic.characteristicId, Uuid([01]));
+      });
+
+      test('it converts value', () {
+        final result = sut.characteristicValueFrom(message);
+        final charValue = result.result.iif(
+            success: (v) => v,
+            failure: (_) => throw AssertionError("Not expected to fail"));
+        expect(charValue, value);
```
werediver: Why not `expect(charValue, Result.success(v))`.

test/protobuf_converter_test.dart
```diff
@@ -183,6 +186,204 @@ void main() {
...
+
+    group("Converts writecharacteristic info", () {
+      const id = 'id';
+      final characteristic = pb.CharacteristicAddress()
+        ..deviceId = id
+        ..serviceUuid = (pb.Uuid()..data = [0])
+        ..characteristicUuid = (pb.Uuid()..data = [1]);
+
+      final message = pb.WriteCharacteristicInfo()
+        ..characteristic = characteristic;
```
werediver: Should it be in a set-up block?

werediver:
Approved, but please check the latest comments :)

Issue #18, opened 2020-02-20
Remove global set error handler for Rxjava


remonh87:

werediver:
Resolved by #26.

PR #26, opened 2020-02-24
remove rxjava errorhandler

remonh87:
people that use library themselves should implement custom rx handler. Since it is a singleton we do not want to possibly override customhandlers


remonh87:
Fixes #18

werediver:
But shall we put this one into the 2.0 milestone?

remonh87:
ah yes good one I did add the issue to the milestone but makes sense to add pr as well

PR #29, opened 2020-02-25
Add disconnect from device doc with example

### LINKED ISSUE #27: Ability to disconnect from a device
So I was digging in the code looking for a way to programmatically disconnect from a device and found the `DisconnectFromDeviceRequest`in the generated pb dart file. But it seems like this feature is not made available in the Method channel in any way? I would propose to add this as a feature making it possible to also disconnect from a device in the same way you're able to `_ble.connectToDevice(...`

Taym95:
Fix #27

### DIFF
--- a/README.md
+++ b/README.md
@@ -64,6 +64,43 @@ issue use the method `connectToAdvertisingDevice` to first scan for the device a
 ```
 Besides the normal connection parameters that are described at above this function also has 2 additional required parameters: `withService` and  `prescanDuration`. PreScanDuration is the amount of time the ble stack will scan for the device before it attempts to connect (if the device is found)
 
+### Disconnect from device
+
+When you connect to a device with `connectToDevice(⋯)` and you subscribe to a characteristic you get a `StreamSubscription` Keep a subscription to this StreamSubscription as long as you need the connection. By closing the StreamSubscription you terminate the connection.
+
+```dart
+// connect
+void _connectToDevice() {
+  Stream<ConnectionStateUpdate> connectedDeviceStream = reactiveBleClient
+        .connectToDevice(
+          id: 'AA:BB:CC:DD:EE:FF',
+          connectionTimeout: const Duration(seconds: 10),
+        )
+        .asBroadcastStream();
+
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -64,6 +64,43 @@ issue use the method `connectToAdvertisingDevice` to first scan for the device a
 ```
 Besides the normal connection parameters that are described at above this function also has 2 additional required parameters: `withService` and  `prescanDuration`. PreScanDuration is the amount of time the ble stack will scan for the device before it attempts to connect (if the device is found)
 
+### Disconnect from device
```
werediver: This is not the way to document API. Please start with adding a doc-comment to [connectToDevice(⋯)](https://github.com/PhilipsHue/flutter_reactive_ble/blob/adbc381fcb115373de51fcb426596ff508f81a5e/lib/src/reactive_ble.dart#L273). Then—as you in fact did—consider extending the example app. Only if after all of this you still think it's not enough, consider extending the readme.

werediver: This is not the way to document API. Please start with adding a doc-comment to [connectToDevice(⋯)](https://github.com/PhilipsHue/flutter_reactive_ble/blob/adbc381fcb115373de51fcb426596ff508f81a5e/lib/src/reactive_ble.dart#L273). Then—as you in fact did—consider extending the example app. Only if after all of this you still think it's not enough, consider extending the readme.

werediver:
Look, on pub.dev some nice (or at least standardized) docs are generated from our doc-comments. There is a page for [connectToDevice(⋯)](https://pub.dev/documentation/flutter_reactive_ble/latest/flutter_reactive_ble/FlutterReactiveBle/connectToDevice.html), and it's empty (the method implementation really doesn't count). Using the doc-comments we make to documentation available in IDEs, at pub.dev, and in the source code (obviously). Using the readme we force users to read looong text files. Not cool, is it?

werediver:
@Taym95 Are you planning to rework this pull request or somebody take it over?

Taym95:
> @Taym95 Are you planning to rework this pull request or somebody take it over?

Yes, I will rework it.

PR #32, opened 2020-02-27
Add manufacturer data in DiscoveredDevice

### LINKED ISSUE #13: Add Manufacturer specific data in DiscoveredDevice
Could the response of `scanForDevices` contains the Manufacturer specific data?

xvrh:
Fixes #13 

This adds a `Map<int, Uint8List> manufacturerData` field in the `DiscoveredDevice` class.

This exposes the `getManufacturerSpecificData` from Android `ScanRecord` (https://developer.android.com/reference/android/bluetooth/le/ScanRecord#getManufacturerSpecificData())
and the `CBAdvertisementDataManufacturerDataKey` on iOS (https://developer.apple.com/documentation/corebluetooth/cbcentralmanagerdelegate/advertisement_data_retrieval_keys)


android/src/main/kotlin/com/signify/hue/flutterreactiveble/ble/BleWrapper.kt
```diff
@@ -3,7 +3,7 @@ package com.signify.hue.flutterreactiveble.ble
 import com.polidea.rxandroidble2.RxBleConnection
 import java.util.UUID
 
-data class ScanInfo(val deviceId: String, val name: String, val rssi: Int, val serviceData: Map<UUID, ByteArray>)
+data class ScanInfo(val deviceId: String, val name: String, val rssi: Int, val serviceData: Map<UUID, ByteArray>, val manufacturerData: Map<Int, ByteArray>)
```
xvrh: For the record, I'm not really happy about bloating even more this intermediate structure.
We may want to consider just passing around the raw advertised bytes and have an "advertisement parser" on the Dart side.

remonh87: Yess indeed we should just parse the raw advertisement data. Let's create an issue for it

werediver: Which issue? Let's not integrate what creates an issue.

I checked the Manufacturer Specific Data format and I do not understand, how a map is a suitable representation for it. According to the spec it's just a uint16 manufacturer identifier and arbitrary binary payload following. Let's stick to this exact representation.

xvrh: Technically there could have multiple manufacturer specific data sequences in a single advertisement packet.
The iOS API only expose one (and the API doesn't split automatically the identifier code with the payload).

But the Android API (and the Web Bluetooth API) exposes this as a Map<int, bytes>.

I can switch to the iOS model and just do the concatenation on Android.

My comment here was more about the way we are passing data around in Android side with an useless temporary conversion.

werediver: I didn't realize that Android uses `Map<int, bytes>` representation. We can use it, as it's more capable and accommodates the iOS data as well. Is a _map_ here necessary, though? What is the meaning of a key in that map? (if it's just an index, we can use a list; I don't think we should concatenate the data)

werediver: From the Android docs you @xvrh have referred to it looks like the actual type for the aggregate manufacturer specific data is `SparseArray<byte[]>`. Could you clarify why do you suggest using a map, instead of a list for it?

xvrh: The key is the Company Identifier code (https://www.bluetooth.com/specifications/assigned-numbers/company-identifiers/)
This is the first 2 bytes of the manufacturer data sequence.

We can also a `List<Uint8List>` and the user will be responsible to split the company identifier and the payload.
On iOS it will have maximum one entry in the list.

On Android, we would have to loop over the entry in manufacturerSpecificData and recreate the raw payload:
```kotlin
for (i in 0..manufacturerSpecificData.size()) {
    val companyId = manufacturerSpecificData.keyAt(i)
    val payload = manufacturerSpecificData.get(companyId);

    val rawData = mutableListOf<Byte>()
    rawData.add((companyId.toByte()))
    rawData.add(((companyId.shr(8)).toByte()))
    rawData.addAll(2, payload.asList());

    datas.add(rawData);
}
```

I'm fine with all the options

werediver: > The key is the Company Identifier code

Does the spec guarantee that multiple manufacturer specific data sections never share the same manufacturer identifier?

xvrh: @werediver despite its name, `SparseArray` is actually the equivalent of a `Map` with integer keys :-)
https://developer.android.com/reference/android/util/SparseArray

xvrh: > Does the spec guarantee that multiple manufacturer specific data sections never share the same manufacturer identifier?
I'm really not qualified to answer that. 
I guess any device can advertise whatever he wants. It just that the phone won't have the api to access the data correctly.

werediver: > I guess any device can advertise whatever he wants. It just that the phone won't have the api to access the data correctly.

I'd reason the same way. But that means we cannot pack multiple manufacturer specific data sections into a map using their manufacturer identifiers as the map keys, right? If the current implementation is doing that, that's a problem.

android/src/main/kotlin/com/signify/hue/flutterreactiveble/utils/SparseArrayExtension.kt
```diff
@@ -0,0 +1,14 @@
...
+
+fun <E> SparseArray<E>.toMap(): Map<Int, E> {
+    val hashMap = HashMap<Int, E>(size())
+    for (i in 0..size()) {
+        val key = keyAt(i)
+        hashMap[key] = get(key);
+    }
+
+    return hashMap;
+}
```
werediver: This conversion of a sparse array to a map looks really fishy to me. Let's try to avoid it and kill the method.

android/src/main/kotlin/com/signify/hue/flutterreactiveble/ble/BleWrapper.kt
xvrh: > If the current implementation is doing that, that's a problem.
The Android Bluetooth API is doing it.
But they also expose the raw bytes, so any one can still access all the data.

On iOS, only one record is exposed anyway.

werediver: @xvrh What would you say to the following proposition? Let's indeed stick to the simplified approach, as in iOS. That must cover for 99% of the cases and we'll have a possibility to change / extend the implementation later, if there is demand.

android/src/main/kotlin/com/signify/hue/flutterreactiveble/utils/SparseArrayExtension.kt
xvrh: Agree!
I see 2 things that we could do:
- Remove the `ScanInfo` data class and just have the `ReactiveBleClient` directly convert the `polidea.ScanResult` to `protobuf.DeviceScanInfo`

- Don't try to interpret the data (ServiceData and ManufacturerSpecifData) on the Android side, just forward the raw advertisment bytes to the Dart side. (Note that it won't work easily on iOS since they don't expose the raw bytes)

android/src/main/kotlin/com/signify/hue/flutterreactiveble/ble/BleWrapper.kt
xvrh: @werediver OK fine with me.
I'll push an update with the user-facing api be `Uint8List` containing a single manufacturer data entry with the identifier code as the first 2 bytes.

android/src/main/kotlin/com/signify/hue/flutterreactiveble/utils/SparseArrayExtension.kt
remonh87: I would personally go for the first result. Actually the reason I created that intermediate layer was for testing purposes but we can work around this

lib/src/model/discovered_device.dart
```diff
@@ -27,13 +27,19 @@ class DiscoveredDevice extends $DiscoveredDevice {
   @CustomEquality(DeepCollectionEquality())
   final Map<Uuid, Uint8List> serviceData;
 
+  /// Returns a map of Company Identifier Codes and the corresponding manufacturer
+  /// specific data.
+  @CustomEquality(DeepCollectionEquality())
```
remonh87: @werediver actually here we are touching an interesting point. Do we consider a device with different servicedata and / or different manufacturerdata but with same id (macadress on android, uuid iOS) a new device or the same.  I think I can advocate for both cases

android/src/main/kotlin/com/signify/hue/flutterreactiveble/utils/SparseArrayExtension.kt
werediver: @remonh87 If you say so, let's try it. I hope it doesn't push the protobuf data types to deep in the hierarchy of components. I'd prefer to use them near the IPC facilities only.

remonh87: that is a fair point indeed

lib/src/model/discovered_device.dart
werediver: @remonh87 I do not agree with the current implementation at all. We should not abuse the equality test operator to perform some particular "matching", as in ID comparison or approximate comparison or anything. We shouldn't ignore fields at all (we do now).

android/src/main/kotlin/com/signify/hue/flutterreactiveble/utils/SparseArrayExtension.kt
werediver: @remonh87 So, it does push it too deep?

@xvrh Then, please, keep this intermediate structure at least for now. Let's address that question outside of this pull request.

xvrh: @werediver Noted!
I'm just bothered by a kotlin warning when I try to add a `ByteArray` to this class (`Array property in data class: it is recommended to override equals`)

werediver: @xvrh But that warning suggests a solution. Could you implement it?

xvrh: Sure.

xvrh:
Hello,
Sorry for the delay.
I pushed the new version that only expose a single `Uint8List` in the user-facing API.
This is what iOS gives us. 
On Android, we extract the data from the `SparseArray` and recreate the raw manufacturer data entry (first 2 bytes are the manufacturer identifier).

Please let me know what you think.

android/src/main/kotlin/com/signify/hue/flutterreactiveble/ble/BleWrapper.kt
```diff
@@ -3,7 +3,31 @@ package com.signify.hue.flutterreactiveble.ble
 import com.polidea.rxandroidble2.RxBleConnection
 import java.util.UUID
 
-data class ScanInfo(val deviceId: String, val name: String, val rssi: Int, val serviceData: Map<UUID, ByteArray>)
+data class ScanInfo(val deviceId: String, val name: String, val rssi: Int, val serviceData: Map<UUID, ByteArray>, val manufacturerData: ByteArray) {
+    override fun equals(other: Any?): Boolean {
+        if (this === other) return true
+        if (javaClass != other?.javaClass) return false
+
+        other as ScanInfo
```
werediver: Won't this blow, if `other` is not `ScanInfo`? Is that a convention?

I'd expect `1 == "text"` to be false or rejected at compile time.

android/src/main/kotlin/com/signify/hue/flutterreactiveble/ble/BleWrapper.kt
```diff
@@ -3,7 +3,31 @@ package com.signify.hue.flutterreactiveble.ble
...
+
+        if (deviceId != other.deviceId) return false
+        if (name != other.name) return false
+        if (rssi != other.rssi) return false
+        if (serviceData != other.serviceData) return false
+        if (!manufacturerData.contentEquals(other.manufacturerData)) return false
```
werediver: Why not to merge all the expressions with `&&` (logical "and")? It'd be shorter.

android/src/main/kotlin/com/signify/hue/flutterreactiveble/ble/BleWrapper.kt
xvrh: This code was generated by IntelliJ, I will have a look.

ios/Classes/ReactiveBle/Central.swift
```diff
@@ -11,7 +11,7 @@ typealias ServiceData = [ServiceID: Data]
 final class Central {
 
     typealias StateChangeHandler = (Central, CBManagerState) -> Void
-    typealias DiscoveryHandler = (Central, CBPeripheral, ServiceData, RSSI) -> Void
+    typealias DiscoveryHandler = (Central, CBPeripheral, [String: Any], RSSI) -> Void
```
werediver: Is `ServiceData` type-alias still used? If not, let's kill it.

Now it would make sense to define `typealias AdvertisementData = [String: Any]`, I guess. I does not make the program more type-safe, but it does make declarations like this more readable / self-explanatory.

ios/Classes/ReactiveBle/CentralManagerDelegate.swift
```diff
@@ -9,7 +9,7 @@ enum ConnectionChange {
 final class CentralManagerDelegate: NSObject, CBCentralManagerDelegate {
 
     typealias StateChangeHandler = (CBManagerState) -> Void
-    typealias DiscoveryHandler = (CBPeripheral, ServiceData, RSSI) -> Void
+    typealias DiscoveryHandler = (CBPeripheral, [String : Any], RSSI) -> Void
```
werediver: If the type-alias `AdvertisementData` is introduced, please use it here.

android/src/main/kotlin/com/signify/hue/flutterreactiveble/ble/BleWrapper.kt
xvrh: Line 8 checks that other is the same class.
The Kotlin linter warns about a "Array property in data class: it's recommended to override equals() / hashCode()" and provide this recommended implementation.

lib/src/model/discovered_device.dart
```diff
@@ -27,13 +27,18 @@ class DiscoveredDevice extends $DiscoveredDevice {
   @CustomEquality(DeepCollectionEquality())
   final Map<Uuid, Uint8List> serviceData;
 
+  /// Returns the manufacturer specific data part of the advertisement packet.
+  /// The first 2 bytes are the Company Identifier Codes.
+  final Uint8List manufacturerData;
```
werediver: It's not a function, so it doesn't "return" it, it contains it. Maybe, just like this?

```dart
/// Manufacturer specific data. The first 2 bytes are the Company Identifier Codes.
```

android/src/main/kotlin/com/signify/hue/flutterreactiveble/ble/BleWrapper.kt
werediver: Ah, it's generated code. I guess, we can trust IDEA 🙂 Thanks for checking.

android/src/main/kotlin/com/signify/hue/flutterreactiveble/ble/BleWrapper.kt
werediver: As this method is generated , let's keep it as is.

ios/Classes/ReactiveBle/Central.swift
xvrh: Done.

ios/Classes/ReactiveBle/CentralManagerDelegate.swift
xvrh: Done.

lib/src/model/discovered_device.dart
xvrh: Done. Thanks

ios/Classes/ReactiveBle/CentralManagerDelegate.swift
werediver: Ehm. Did you push it?

werediver:
> Hello,
> Sorry for the delay.
> I pushed the new version that only expose a single `Uint8List` in the user-facing API.
> This is what iOS gives us.
> On Android, we extract the data from the `SparseArray` and recreate the raw manufacturer data entry (first 2 bytes are the manufacturer identifier).
> 
> Please let me know what you think.

No worries, thanks for your great contribution and collaboration. I like how it looks now 👍

Let's wait while @remonh87 takes the final look at the pull request, and we can integrate it.

In the meanwhile, could you check [this remark](https://github.com/PhilipsHue/flutter_reactive_ble/pull/32#discussion_r386286598) once more? (or clarify, why not)

werediver:
I'd also suggest to squash all the commits into one before or during the merge, because they start looking a bit stochastic.

ios/Classes/ReactiveBle/CentralManagerDelegate.swift
xvrh: OK, sorry about that. Missed that there were 2 DiscoveryHandler.
Also, I reintroduced ServiceData that was actually used in PluginController

xvrh:
@werediver Squash done.
FYI, you can also squash from github when merging the PR (if you don't see it, it needs to be enabled in the setting then).

werediver:
> @werediver Squash done.
> FYI, you can also squash from github when merging the PR (if you don't see it, it needs to be enabled in the setting then).

That's what I meant by "or during the merge" 😄 Not sure about whether it's enabled in the settings, but that I'll check.

xvrh:
I replaced magic number 8 with a constant `Byte.SIZE_BITS`

PR #34, opened 2020-02-27
Remove "Device found" log

xvrh:
I find this log to be very spammy.
Please consider either removing it (this pull request) or provide (or document) a way to configure the log level.

### DIFF
--- a/android/src/main/kotlin/com/signify/hue/flutterreactiveble/channelhandlers/ScanDevicesHandler.kt
+++ b/android/src/main/kotlin/com/signify/hue/flutterreactiveble/channelhandlers/ScanDevicesHandler.kt
@@ -38,7 +38,6 @@ class ScanDevicesHandler(private val bleClient: com.signify.hue.flutterreactiveb
                         .observeOn(AndroidSchedulers.mainThread())
                         .subscribe(
                         { scanResult ->
-                            Timber.d("Device found ${scanResult.deviceId} ${scanResult.name}")
                             handleDeviceScanResult(converter.convertScanInfo(scanResult))
                         },
                         { throwable ->


werediver:
This looks completely fine to me. As long as @remonh87 does not have a good reason to keep that logging, I'd like to integrate it.

android/src/main/kotlin/com/signify/hue/flutterreactiveble/channelhandlers/ScanDevicesHandler.kt
```diff
@@ -38,7 +38,6 @@ class ScanDevicesHandler(private val bleClient: com.signify.hue.flutterreactiveb
                         .observeOn(AndroidSchedulers.mainThread())
                         .subscribe(
                         { scanResult ->
-                            Timber.d("Device found ${scanResult.deviceId} ${scanResult.name}")
```
remonh87: This log is  indeed bit intrusive but it would only happen during debug builds. So I think it can be still handy for debugging purposes (to check if on native level the scan results are shown).

xvrh: Can I run a flutter app in normal debug mode but have the android code compiled in release?

werediver: It might be useful for "flutter_reactive_ble" debugging, but much less for an app debugging, given the package works fine. And during the app debugging this line is going to pollute the log, I fully agree with @xvrh. The debugging output should be absent or opt-in. The iOS implementation doesn't log anything, if I remember correctly.

werediver: > Can I run a flutter app in normal debug mode but have the android code compiled in release?

I doubt. At least this is not a standard feature of the toolchain, as far as I'm aware.

remonh87: maybe let's remove it for now and I will create a advanced logging feature. It helped me a lot in the pass but can live with it

werediver: @remonh87 You can always very easily add it _in your working copy_ temporarily, if needed.

Thank you for taking a look.

Issue #35, opened 2020-02-27
Improve architecture example app


remonh87:
**Is your feature request related to a problem? (please describe)**
The architecture does not reflect a super nice way how to user our library in the app

**Describe the solution you'd like**
Extract BLE related code into a seperate component (e.g. BLOC)

werediver:
Thanks @remonh87 for completing that great redesign in #38 🎉

If we have more ideas on this topic, we can track them with new tickets.

PR #36, opened 2020-02-27
Minor fixes in the example app + show live RSSI during scanning

werediver:


example/lib/src/ui/device_list.dart
```diff
@@ -70,19 +68,24 @@ class _DeviceListState extends State<DeviceList> {
...
           if (mounted) {
             setState(() {
-              _devices.add(device);
+              final knownDeviceIndex =
+                  _devices.indexWhere((d) => d.id == device.id);
+
+              if (knownDeviceIndex >= 0) {
+                _devices[knownDeviceIndex] = device;
+              } else {
+                log("New device found: $device");
```
werediver: Honestly, I'd kill all the logging in the example app.

remonh87: kill it

werediver: As discussed via a side-channel, let's kill it during the course of #35.

PR #37, opened 2020-02-28
Update "README.md"

werediver:


README.md
```diff
@@ -1,9 +1,13 @@
 # Flutter reactive BLE library
 
+[![flutter_reactive_ble version](https://img.shields.io/pub/v/flutter_reactive_ble?label=flutter_reactive_ble)](https://pub.dev/packages/flutter_reactive_ble)
+
```
werediver: A fancy shield showing the latest version of the package on pub.dev.

werediver:
The minor changes (extra empty lines for instance) are according to the Markdown linter of VS Code.

README.md
```diff
@@ -25,74 +30,78 @@ reactivebleclient.scanForDevices(withService: uuid, scanMode: ScanMode.lowLatenc
...
-If no scanmode is specified the library will use the balanced scanmode as reference.
+The `withService` parameter is required.  The parameter `scanMode` is only used on Android and follows the conventions described on [ScanSettings](https://developer.android.com/reference/android/bluetooth/le/ScanSettings#SCAN_MODE_BALANCED) Android reference page. If `scanMode` is omitted the balanced scan mode will be used.
 
 ### Establishing connection
 
 To interact with a device you first need to establish a connection:
 
 ```dart
 reactiveBleClient.connectToDevice(
-      id: 'AA:BB:CC:DD:EE:FF',
```
werediver: I removed all device IDs like this, because that suggest you can type something in. No, you can only get a valid device ID by performing a scan.

README.md
```diff
@@ -25,74 +30,78 @@ reactivebleclient.scanForDevices(withService: uuid, scanMode: ScanMode.lowLatenc
...
 ```dart
-final characteristic = QualifiedCharacteristic(serviceId: serviceUuid, characteristicId: characteristicUuid, deviceId: 'AA:BB:CC:DD:EE:FF');
+final characteristic = QualifiedCharacteristic(serviceId: serviceUuid, characteristicId: characteristicUuid, deviceId: foundDeviceId);
 final response = await reactiveBleClient.readCharacteristic(characteristic);
 ```
 
 #### Write with response
-Write a value to characteristic and await the reaponse. The return value is a `List<Int>`.
+
+Write a value to characteristic and await the response. The return value is a list of bytes (`List<int>`).
```
werediver: Actually, I think we should switch to `Uint8List`, because it suggests it means exactly "list of bytes", whereas `List<int>` is ambiguous. `Uint8List` is also a `TypedData`, which is useful when decoding blobs.

README.md
remonh87: makes sense is more clear what we mean indeed

PR #41, opened 2020-03-06
Support blescan with empty filter

### LINKED ISSUE #12: Allow scanForDevices without service uuid
The README for `scanForDevices` states that "the `withService` parameter is required.".

I would like to scan for every BLE devices around. What is the reason to enforce the service filter?

### LINKED ISSUE #23: Can't scan with location service disabled
Hi,
I'm unable to scan for devices when the location services are disabled:

```D/ScanDevicesHandler$startDeviceScan$$inlined$let$lambda(26938): Error while scanning for devices:  Location Services disabled (code 4)```

I'm not sure what I'm doing wrong but I think I should be able to scan for devices when only the bluetoothservices are enabled.

I'm getting this on Android 8, with flutter_reactive_blue 1.1.0

I'm just doing something along the line of:

```dart
_scanSubscription = bleClient.scanForDevices(
  withService: BTDevice.serviceUuid,
  scanMode: ScanMode.lowLatency,
).listen(
  (device) {
  },
);
```

I also added all the permissions required in the AndroidManifest.xml:

```xml
    <uses-permission android:name="android.permission.INTERNET" />
    <uses-permission android:name="android.permission.CAMERA" />
    <uses-permission android:name="android.permission.ACCESS_FINE_LOCATION" />
    <uses-permission android:name="android.permission.BLUETOOTH" />
    <uses-permission android:name="android.permission.BLUETOOTH_ADMIN" />
```

By the way, using FlutterBlue on the same device doesn't require location services to be on in the phone settins.

remonh87:
Fixes #12 
Fixes #23

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,7 +22,7 @@ The reactive BLE lib supports the following:
 Discovering BLE devices should be done like this:
 
 ```dart
-reactivebleclient.scanForDevices(withServices: [uuid], scanMode: ScanMode.lowLatency).listen((device) {
+reactivebleclient.scanForDevices(withServices: [serviceId], scanMode: ScanMode.lowLatency).listen((device) {
       //code for handling results
     }, onError: () {
       //code for handling error
@@ -30,7 +30,7 @@ reactivebleclient.scanForDevices(withServices: [uuid], scanMode: ScanMode.lowLat
   
 ```
 
-The `withServices` parameter is required if you do not want to filter on services pass an empty list.  The parameter `scanMode` is only used on Android and follows the conventions described on [ScanSettings](https://developer.android.com/reference/android/bluetooth/le/ScanSettings#SCAN_MODE_BALANCED) Android reference page. If `scanMode` is omitted the balanced scan mode will be used.
+The `withServices` parameter specifies the advertised service IDs to look for. If an empty list is passed, all the advertising devices will be reported. The parameter `scanMode` is only used on Android and follows the conventions described on [ScanSettings](https://developer.android.com/reference/android/bluetooth/le/ScanSettings#SCAN_MODE_BALANCED) Android reference page. If `scanMode` is omitted the balanced scan mode will be used.
 
 ### Establishing connection
 
--- a/README.md
+++ b/README.md
@@ -22,7 +22,7 @@ The reactive BLE lib supports the following:
 Discovering BLE devices should be done like this:
 
 ```dart
-reactivebleclient.scanForDevices(withService: uuid, scanMode: ScanMode.lowLatency).listen((device) {
+reactivebleclient.scanForDevices(withServices: [uuid], scanMode: ScanMode.lowLatency).listen((device) {
       //code for handling results
     }, onError: () {
       //code for handling error
@@ -30,7 +30,7 @@ reactivebleclient.scanForDevices(withService: uuid, scanMode: ScanMode.lowLatenc
   
 ```
 
-The `withService` parameter is required.  The parameter `scanMode` is only used on Android and follows the conventions described on [ScanSettings](https://developer.android.com/reference/android/bluetooth/le/ScanSettings#SCAN_MODE_BALANCED) Android reference page. If `scanMode` is omitted the balanced scan mode will be used.
+The `withServices` parameter is required if you do not want to filter on services pass an empty list.  The parameter `scanMode` is only used on Android and follows the conventions described on [ScanSettings](https://developer.android.com/reference/android/bluetooth/le/ScanSettings#SCAN_MODE_BALANCED) Android reference page. If `scanMode` is omitted the balanced scan mode will be used.
 
 ### Establishing connection
 
--- a/android/src/main/kotlin/com/signify/hue/flutterreactiveble/channelhandlers/ScanDevicesHandler.kt
+++ b/android/src/main/kotlin/com/signify/hue/flutterreactiveble/channelhandlers/ScanDevicesHandler.kt
@@ -19,7 +19,6 @@ class ScanDevicesHandler(private val bleClient: com.signify.hue.flutterreactiveb
     private var scanParameters: ScanParameters? = null
 
     override fun onListen(objectSink: Any?, eventSink: EventChannel.EventSink?) {
-
         eventSink?.let {
             scanDevicesSink = eventSink
             startDeviceScan()
@@ -27,7 +26,7 @@ class ScanDevicesHandler(private val bleClient: com.signify.hue.flutterreactiveb
     }
 
     override fun onCancel(objectSink: Any?) {
-        Timber.d("Scanfordevices cancelled called")
+        Timber.d("Scanning canceled")
         stopDeviceScan()
         scanDevicesSink = null
     }
@@ -41,13 +40,12 @@ class ScanDevicesHandler(private val bleClient: com.signify.hue.flutterreactiveb
                                 handleDeviceScanResult(converter.convertScanInfo(scanResult))
                             },
                             { throwable ->
-
-                                Timber.d("Error while scanning for devices:  ${throwable.message}")
+                                Timber.d("Error while scanning for devices: ${throwable.message}")
                                 handleDeviceScanResult(converter.convertScanErrorInfo(throwable.message))
                             }
                     )
         }
-                ?: handleDeviceScanResult(converter.convertScanErrorInfo("Scanparameters are not being set"))
+                ?: handleDeviceScanResult(converter.convertScanErrorInfo("Scanning parameters are not set"))
     }
 
     fun stopDeviceScan() {
--- a/android/src/main/kotlin/com/signify/hue/flutterreactiveble/channelhandlers/ScanDevicesHandler.kt
+++ b/android/src/main/kotlin/com/signify/hue/flutterreactiveble/channelhandlers/ScanDevicesHandler.kt
@@ -22,7 +22,7 @@ class ScanDevicesHandler(private val bleClient: com.signify.hue.flutterreactiveb
 
         eventSink?.let {
             scanDevicesSink = eventSink
-                startDeviceScan()
+            startDeviceScan()
         }
     }
 
@@ -33,19 +33,21 @@ class ScanDevicesHandler(private val bleClient: com.signify.hue.flutterreactiveb
     }
 
     private fun startDeviceScan() {
-            scanParameters?.let { params ->
-                scanForDevicesDisposable = bleClient.scanForDevices(params.filter, params.mode)
-                        .observeOn(AndroidSchedulers.mainThread())
-                        .subscribe(
-                        { scanResult ->
-                            handleDeviceScanResult(converter.convertScanInfo(scanResult))
-                        },
-                        { throwable ->
-                            Timber.d("Error while scanning for devices:  ${throwable.message}")
-                            handleDeviceScanResult(converter.convertScanErrorInfo(throwable.message))
-                        }
-                )
-            } ?: handleDeviceScanResult(converter.convertScanErrorInfo("Scanparameters are not being set"))
+        scanParameters?.let { params ->
+            scanForDevicesDisposable = bleClient.scanForDevices(params.filter, params.mode, params.locationServiceIsMandatory)
+                    .observeOn(AndroidSchedulers.mainThread())
+                    .subscribe(
+                            { scanResult ->
+                                handleDeviceScanResult(converter.convertScanInfo(scanResult))
+                            },
+                            { throwable ->
+
+                                Timber.d("Error while scanning for devices:  ${throwable.message}")
+                                handleDeviceScanResult(converter.convertScanErrorInfo(throwable.message))
+                            }
+                    )
+        }
+                ?: handleDeviceScanResult(converter.convertScanErrorInfo("Scanparameters are not being set"))
     }
 
     fun stopDeviceScan() {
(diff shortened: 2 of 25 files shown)


lib/src/reactive_ble.dart
```diff
@@ -220,14 +220,22 @@ class FlutterReactiveBle {
...
-      {@required Uuid withService, ScanMode scanMode = ScanMode.balanced}) {
+  /// Scan for devices that are advertising a specific services. Use [withServices] filter on devices that are advertising
+  /// the defined services. 
+  /// 
+  /// There are two Android specific parameters that will be ignored on iOS: 
+  /// [scanMode] is used to enforce a more battery or latency intensive scan strategy,
+  /// [locationServiceIsMandatory] is used to enforce location services to be enabled while scanning.
+  /// It is recommended to set this to true when supporting Android 7 and lower.
+  Stream<DiscoveredDevice> scanForDevices({
+    @required List<Uuid> withServices ,
```
werediver: Doesn't VS Code autoformat Dart sources for you?

remonh87:
Update readme with correct example

lib/src/reactive_ble.dart
```diff
@@ -220,14 +220,22 @@ class FlutterReactiveBle {
...
 
-  /// Scan for devices that are advertising a specific service
-  /// [scanMode] is used only on Android to enforce a more battery or latency intensive scan strategy
-  ///
-  Stream<DiscoveredDevice> scanForDevices(
-      {@required Uuid withService, ScanMode scanMode = ScanMode.balanced}) {
+  /// Scan for devices that are advertising a specific services. Use [withServices] filter on devices that are advertising
+  /// the defined services. 
+  /// 
+  /// There are two Android specific parameters that will be ignored on iOS:
```
werediver: I'd suggest "are ignored".

lib/src/reactive_ble.dart
```diff
@@ -220,14 +220,22 @@ class FlutterReactiveBle {
...
-  /// Scan for devices that are advertising a specific service
-  /// [scanMode] is used only on Android to enforce a more battery or latency intensive scan strategy
-  ///
-  Stream<DiscoveredDevice> scanForDevices(
-      {@required Uuid withService, ScanMode scanMode = ScanMode.balanced}) {
+  /// Scan for devices that are advertising a specific services. Use [withServices] filter on devices that are advertising
+  /// the defined services. 
+  /// 
+  /// There are two Android specific parameters that will be ignored on iOS: 
+  /// [scanMode] is used to enforce a more battery or latency intensive scan strategy,
```
werediver: Maybe "allows to choose between different levels of power efficient and/or low latency scan modes"?

lib/src/reactive_ble.dart
```diff
@@ -220,14 +220,22 @@ class FlutterReactiveBle {
...
+  /// [locationServiceIsMandatory] is used to enforce location services to be enabled while scanning.
```
werediver: What means "to enforce"? Will the library turn in off?

lib/src/reactive_ble.dart
```diff
@@ -220,14 +220,22 @@ class FlutterReactiveBle {
...
+  /// It is recommended to set this to true when supporting Android 7 and lower.
```
werediver: Does the library support Android versions lower than 7?

protos/Readme.md
```diff
@@ -5,12 +5,10 @@
...
 ```brew isntall dart```
-4. Get the latest dart-protoc-plugin from https://github.com/dart-lang/dart-protoc-plugin
-5. Go to dart-protoc-plugin-<<version>> folder and call
-```pub install```
-6. Add plugin path to PATH
-7. Make sure protoc + protoc-gen-dart + dart bins are all in the same path
-8. Run the following command from the protos folder
+4. Run ```pub global activate protoc_plugin```
+5. OPTIONAL Add plugin path to PATH
+6. Make sure protoc + protoc-gen-dart + dart bins are all in the same path
```
werediver: I'd suggest using commas instead of pluses. Also, please reword the ending. What means "in the same path"? I'm pretty sure it's not required to have those tools in the same folder.

protos/bledata.proto
```diff
@@ -3,8 +3,9 @@ option java_package = "com.signify.hue.flutterreactiveble";
 option java_outer_classname = "ProtobufModel";
 
 message ScanForDevicesRequest {
-    Uuid serviceUuid = 1;
+    repeated Uuid serviceUuids = 1;
     int32 scanMode = 2;
+    bool locationServiceMandatory=3;
```
werediver: When you decide on the right name to use everywhere, use it here as well.

lib/src/reactive_ble.dart
espositofulvio: I'm following this PR very closely because it would allow me to switch to flutter_reactive_ble. Is it possible to lower the minsdk version to at least 21 if there aren't any technical hurdles? It looks like other libraries based on rxandroidble supports android SDK from 18 onwards and would be good to have support for device which are a bit older but still make up for the vast majority of market share

remonh87: When we started with the ble library we decided from beginning to support 4 Android versions (7,8,9 and 10). Reason is the fragmentation on Android and the fact BLE behaves totally different on Android versions below 7. We prefer to be good on a more limited range of Android versions. Of course if there is more demand from the community and ppl are willing to help us out we can consider supporting more Android versions.

protos/Readme.md
remonh87: this is not needed left over from the old times removed it :)

remonh87:
Todo fix wrong status propagation BluetoothDiscoveryDisabled when location is turned off This is wrong and we should only check if the permission is granted

remonh87:
@werediver  thanks for the review fixed the comments. Prefer to do exception handling regarding scanning in a different pr. Did some testing but found some peculiarities that I would like to discuss with you offline.

werediver:
> @werediver thanks for the review fixed the comments. Prefer to do exception handling regarding scanning in a different pr. Did some testing but found some peculiarities that I would like to discuss with you offline.

Thanks for your work! I will review the changes on Monday.

README.md
```diff
@@ -22,15 +22,15 @@ The reactive BLE lib supports the following:
...
+reactivebleclient.scanForDevices(withServices: [uuid], scanMode: ScanMode.lowLatency).listen((device) {
       //code for handling results
     }, onError: () {
       //code for handling error
     });
   
 ```
 
-The `withService` parameter is required.  The parameter `scanMode` is only used on Android and follows the conventions described on [ScanSettings](https://developer.android.com/reference/android/bluetooth/le/ScanSettings#SCAN_MODE_BALANCED) Android reference page. If `scanMode` is omitted the balanced scan mode will be used.
+The `withServices` parameter is required if you do not want to filter on services pass an empty list.  The parameter `scanMode` is only used on Android and follows the conventions described on [ScanSettings](https://developer.android.com/reference/android/bluetooth/le/ScanSettings#SCAN_MODE_BALANCED) Android reference page. If `scanMode` is omitted the balanced scan mode will be used.
```
werediver: This is a single sentence:

> The `withServices` parameter is required if you do not want to filter on services pass an empty list.

It seems broken to me. One way to fix it is like this: "The `withServices` parameter specifies the advertised service IDs to look for. If an empty list is passed, all the advertising devices will be reported."

android/src/main/kotlin/com/signify/hue/flutterreactiveble/channelhandlers/ScanDevicesHandler.kt
```diff
@@ -33,19 +33,21 @@ class ScanDevicesHandler(private val bleClient: com.signify.hue.flutterreactiveb
...
-            } ?: handleDeviceScanResult(converter.convertScanErrorInfo("Scanparameters are not being set"))
+        scanParameters?.let { params ->
+            scanForDevicesDisposable = bleClient.scanForDevices(params.filter, params.mode, params.locationServiceIsMandatory)
+                    .observeOn(AndroidSchedulers.mainThread())
+                    .subscribe(
+                            { scanResult ->
+                                handleDeviceScanResult(converter.convertScanInfo(scanResult))
+                            },
+                            { throwable ->
+
```
werediver: Why an empty line here?

android/src/main/kotlin/com/signify/hue/flutterreactiveble/channelhandlers/ScanDevicesHandler.kt
```diff
@@ -33,19 +33,21 @@ class ScanDevicesHandler(private val bleClient: com.signify.hue.flutterreactiveb
...
+                                Timber.d("Error while scanning for devices:  ${throwable.message}")
```
werediver: Why pollute the log? The caller will know about the failure, because the exception is propagated to them (right?).

As I mentioned before, my opinion is that a library should no pollute the log. At all. The library users will not (normally) debug it, therefore such logging will just complicate their life. For us it's not needed either, unless we are specifically debugging library-level a BLE issue, in which case it's not a problem to add the relevant logging.

I understand it was there before. Perhaps, we should clean the logging in the Android implementation up. At least that would be mine proposition.

android/src/main/kotlin/com/signify/hue/flutterreactiveble/channelhandlers/ScanDevicesHandler.kt
```diff
@@ -33,19 +33,21 @@ class ScanDevicesHandler(private val bleClient: com.signify.hue.flutterreactiveb
...
+                                handleDeviceScanResult(converter.convertScanErrorInfo(throwable.message))
+                            }
+                    )
+        }
+                ?: handleDeviceScanResult(converter.convertScanErrorInfo("Scanparameters are not being set"))
```
werediver: What's `Scanparameters`? Casing doesn't much the variable name and neither it is natural language. I'd suggest to use natural language; if you don't find this a good idea, then match the name of the variable exactly, otherwise it's confusing.

example/lib/src/ui/ble_status_screen.dart
```diff
@@ -16,8 +16,8 @@ class BleStatusScreen extends StatelessWidget {
         return "Authorize the FlutterReactiveBle example app to use Bluetooth and location";
       case BleStatus.poweredOff:
         return "Bluetooth is powered off on your device turn it on";
-      case BleStatus.discoveryDisabled:
-        return "Bluetooth discovery is disabled turn on location services";
+      case BleStatus.locationServicesDisabled:
```
werediver: Nice renaming! 👍

example/lib/src/ui/device_list.dart
```diff
@@ -53,17 +53,21 @@ class _DeviceListState extends State<_DeviceList> {
...
+        return false;
+      }
     }
   }
 
   void _startScanning() {
-    final uuid = Uuid.parse(_uuidController.text);
-    widget.startScan(uuid);
+    final text = _uuidController.text;
+    widget.startScan(text.isEmpty ? null : [Uuid.parse(_uuidController.text)]);
```
werediver: If you _are_ using a list here now, the why do you still use `null`?? Just pass an empty list.

lib/src/model/ble_status.dart
```diff
@@ -3,6 +3,6 @@ enum BleStatus {
   unsupported,
   unauthorized,
   poweredOff,
-  discoveryDisabled,
+  locationServicesDisabled,
```
werediver: That's more precise 👍

lib/src/prescan_connector.dart
```diff
@@ -27,22 +28,22 @@ class PrescanConnector {
     Duration connectionTimeout,
   }) connectDevice;
 
-  final Stream<DiscoveredDevice> Function({Uuid withService, ScanMode scanMode})
-      scanDevices;
+  final Stream<DiscoveredDevice> Function(
+      {List<Uuid> withServices, ScanMode scanMode}) scanDevices;
```
werediver: I prefer single parameter per line in such cases.

lib/src/reactive_ble.dart
```diff
@@ -220,14 +220,22 @@ class FlutterReactiveBle {
         .then((message) => message.result.dematerialize());
   }
 
-  /// Scan for devices that are advertising a specific service
-  /// [scanMode] is used only on Android to enforce a more battery or latency intensive scan strategy
-  ///
-  Stream<DiscoveredDevice> scanForDevices(
-      {@required Uuid withService, ScanMode scanMode = ScanMode.balanced}) {
+  /// Scan for devices that are advertising a specific services. Use [withServices] filter on devices that are advertising
```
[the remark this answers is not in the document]
werediver: So, you marked this comment as "resolved". Are you sure? Please, check once more.

lib/src/reactive_ble.dart
```diff
@@ -220,14 +220,22 @@ class FlutterReactiveBle {
...
-  /// Scan for devices that are advertising a specific service
-  /// [scanMode] is used only on Android to enforce a more battery or latency intensive scan strategy
+  /// Scan for devices that are advertising a specific services. Use [withServices] filter on devices that are advertising
+  /// the defined services.
   ///
-  Stream<DiscoveredDevice> scanForDevices(
-      {@required Uuid withService, ScanMode scanMode = ScanMode.balanced}) {
+  /// There are two Android specific parameters that will be ignored on iOS:
+  /// [scanMode] allows to choose between different levels of power efficient and/or low latency scan modes.
+  /// [requireLocationServicesEnabled] is used to enforce location services to be enabled while scanning. If this parameter set to
```
werediver: If you want to break this text into multiple paragraphs, you need to add empty lines to indicate that (are you familiar with Markdown?). If you want to make it a list, you need some bullets (hyphens or asterisks).

lib/src/reactive_ble.dart
```diff
@@ -220,14 +220,22 @@ class FlutterReactiveBle {
...
+  /// true and the location services on the device are disabled it will throw [Exception]. Default is set to true,
```
werediver: Is `[Exception]` here really more informative than "an exception"?

lib/src/reactive_ble.dart
```diff
@@ -220,14 +220,22 @@ class FlutterReactiveBle {
...
+  /// setting the value to false can result in not finding BLE devices on some Android devices.
```
werediver: Can we be more specific and mention Android versions older than 7 or something?

remonh87:
thanks let's ship it 👍

PR #45, opened 2020-03-18
improve dartdoc for reactive_ble

remonh87:
@werediver this is a temp PR to use for discussion on the style of documenting and the amount of information we should provide. 

I was a bit struggling so hopefully you can help out.

### DIFF
--- a/lib/src/reactive_ble.dart
+++ b/lib/src/reactive_ble.dart
@@ -19,7 +19,8 @@ import 'package:flutter_reactive_ble/src/rx_ext/repeater.dart';
 import 'package:flutter_reactive_ble/src/rx_ext/serial_disposable.dart';
 import 'package:meta/meta.dart';
 
-///  Ble client that is single point of entry for ble operations
+/// [FlutterReactiveBle] is the facade of the library. Its interface allows to
+/// perform all the supported BLE operations.
 class FlutterReactiveBle {
   static final FlutterReactiveBle _sharedInstance = FlutterReactiveBle._();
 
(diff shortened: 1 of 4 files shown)


lib/src/reactive_ble.dart
```diff
@@ -28,41 +29,22 @@ class FlutterReactiveBle {
     _trackStatus();
   }
 
-  final _methodChannel = const MethodChannel("flutter_reactive_ble_method");
-
+  /// Registry that keeps track of all BLE devices found during BLE scan.
```
remonh87: I think we shouldn't have exposed this publicly

werediver: This registry is mutable, which allows an API consumer to mess with its contents. From this point of view, yes, it should definitely be private.

lib/src/reactive_ble.dart
```diff
@@ -28,41 +29,22 @@ class FlutterReactiveBle {
...
-    _statusStream.listen((status) => _status = status);
-  }
-
-  final Stream<ScanResult> _scanStream =
-      const EventChannel("flutter_reactive_ble_scan")
-          .receiveBroadcastStream()
-          .cast<List<int>>()
-          .map((data) => pb.DeviceScanInfo.fromBuffer(data))
-          .map(const ProtobufConverter().scanResultFrom);
+  /// Returns the current [BleStatus] of the device.
```
werediver: The use of the word "device" is ambiguous: is it about a BLE peripheral or the host BLE central? We should be careful with this.

lib/src/reactive_ble.dart
```diff
@@ -79,13 +64,43 @@ class FlutterReactiveBle {
...
 
   Future<void> _initialization;
 
   ScanSession _currentScan;
 
+  /// Initializes the BLE client on native level.
+  ///
+  /// Use this method to explicitly initialize the client. It is not required
+  /// to call initialize before starting a ble operation, the [FlutterReactiveBle]
+  /// make sure that it is initialized before executing the operation.
```
werediver: I'd recommend _against_ encouraging people to use both `initializa()` and `deinitialize()` methods. In fact, I'd make them private. Will update the doc-comments to avoid encouragement.

werediver:
I've pushed a commit with my updates. As long as you're good with it, we can proceed and merge it.

PR #46, opened 2020-03-23
Minor fixes

miguelcmedeiros:
In order to add this plugin to https://github.com/flutter/tests I need to fix a test in prescan_connector_test.dart. 
While doing this I added a couple of commits to fix analyzer issues and pub spec.yaml cleanup.

### DIFF
--- a/test/prescan_connector_test.dart
+++ b/test/prescan_connector_test.dart
@@ -239,7 +239,7 @@ void main() {
                           failure: null)
                     ]));
 
-            completer.completeError(null);
+            completer.completeError(Exception());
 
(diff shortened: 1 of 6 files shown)


test/prescan_connector_test.dart
```diff
@@ -239,7 +239,7 @@ void main() {
                           failure: null)
                     ]));
 
-            completer.completeError(null);
+            completer.completeError("");
```
werediver: This is a "no". The official Dart recommendation is to throw either exceptions or errors. Try `Exception()` or alike.

PR #47, opened 2020-03-23
Revert removal of flutter dependency in root pubspec.yaml

miguelcmedeiros:
It seems that I cleaned-up a bit too much in my previous PR, although nothing (pub get, analyzer or tests) hinted that there was an issue.
It seems that we can't remove the flutter dependency from the package, even when it doesn't seem to depend on it... Now I can that it was a wrong assumption: tooling relies on it apparently.

Anyway, this could have been caught by running `flutter pub publish --dry-run`.
@remonh87, maybe you can add a step in the PR qualification pipeline to run this command.

### DIFF
--- a/pubspec.lock
+++ b/pubspec.lock
@@ -163,7 +163,7 @@ packages:
     source: hosted
     version: "0.10.11"
   flutter:
-    dependency: transitive
+    dependency: "direct main"
     description: flutter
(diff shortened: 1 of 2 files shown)


remonh87:
Good one I also changed the pipeline a bit let's see how the build behaves now

Issue #58, opened 2020-04-24
Connecting with connectionTimeout parameter fails immediately with GATT ERROR 133 


vsxed:
**Describe the bug**
Connecting to a device with a connectionTimeout value given, results most of the times in a GATT ERROR 133 (see the logcat log attached below). Not having a connectionTimeout parameter set results in a successful (and pretty fast) connection.

Seems like an android-only issue, the same code works with iOS just fine.

**To Reproduce**
1. Connect to one device using `.connectToDevice(id: ..., connectionTimeout: Duration(seconds: 10))`
2. See the failed attempt to connect in you log

**Expected behavior**
The expected behavior would be an established connection or a timeout after the set duration

- [x] I tried doing the same with a general BLE scanner application

**Smartphone / tablet**
 - Device: OnePlus 5t
 - OS: Android 9
 - Package version: 2.1.0

**Peripheral device**
- Vendor, model: Custom
- Does it run a custom firmware / software: no

**Additional context**
```
E/AccessibilityBridge( 7512): VirtualView node must not be the root node.
D/PluginController( 7512): Start connecting for device 78:DB:2F:D0:BC:D2
D/PluginController( 7512): Start connecting for device 78:DB:2F:D0:BC:D2
D/DeviceConnectionHandler( 7512): Try to connect to device 78:DB:2F:D0:BC:D2
D/DeviceConnectionHandler( 7512): Try to connect to device 78:DB:2F:D0:BC:D2
D/DeviceConnector( 7512): Connecting device 78:DB:2F:D0:BC:D2
D/DeviceConnector( 7512): Connecting device 78:DB:2F:D0:BC:D2
D/DeviceConnectionHandler$listenToConnectionChanges( 7512): Update device: 78:DB:2F:D0:BC:D2 status=0
D/DeviceConnectionHandler$listenToConnectionChanges( 7512): Update device: 78:DB:2F:D0:BC:D2 status=0
D/BluetoothGatt( 7512): connect() - device: 78:DB:2F:D0:BC:D2, auto: false
D/BluetoothGatt( 7512): registerApp()
D/BluetoothGatt( 7512): registerApp() - UUID=d1024742-f90f-4f43-ab6d-1f9fbfe29569
D/BluetoothGatt( 7512): onClientRegistered() - status=0 clientIf=9
I/flutter ( 7512): ### CONNECTING ###
E/AccessibilityBridge( 7512): VirtualView node must not be the root node.
[...]

PR #59, opened 2020-04-28
add additional explanation about effect of no connection timeout

remonh87:
Useage of autoconnect was not clear in the docs.  Issue #58 triggered me on improving the docs

### DIFF
--- a/lib/src/reactive_ble.dart
+++ b/lib/src/reactive_ble.dart
@@ -328,7 +328,8 @@ class FlutterReactiveBle {
   /// this can improve the connection speed on iOS since no full service discovery will be executed. On Android
   /// this variable is ignored since partial discovery is not possible.
   /// The [connectionTimeout] parameter can be used to emit a failure after a certain period in case the connection
-  /// is not established. The pending connection attempt will be cancelled.
+  /// is not established. The pending connection attempt will be cancelled. On Android when no timeout is specified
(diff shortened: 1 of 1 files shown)


lib/src/reactive_ble.dart
```diff
@@ -328,7 +328,8 @@ class FlutterReactiveBle {
   /// this can improve the connection speed on iOS since no full service discovery will be executed. On Android
   /// this variable is ignored since partial discovery is not possible.
   /// The [connectionTimeout] parameter can be used to emit a failure after a certain period in case the connection
-  /// is not established. The pending connection attempt will be cancelled.
+  /// is not established. The pending connection attempt will be cancelled. For Android no specified timeout means
+  /// a connection will be established using autoconnect.
```
werediver, line 332: I guess, it's fine, but my take on this would be a tad different:

> On Android when no timeout is specified the `autoConnect` flag is set in the [connectGatt()](https://developer.android.com/reference/android/bluetooth/BluetoothDevice#connectGatt(android.content.Context,%20boolean,%20android.bluetooth.BluetoothGattCallback)) call, otherwise it's cleared.

Markdown:

```markdown
On Android when no timeout is specified the `autoConnect` flag is set in the [connectGatt()](https://developer.android.com/reference/android/bluetooth/BluetoothDevice#connectGatt(android.content.Context,%20boolean,%20android.bluetooth.BluetoothGattCallback)) call, otherwise it's cleared.```

remonh87: I was doubting to put the link in it already you convinced me 👍

PR #64, opened 2020-05-11
Improve doc comments

### LINKED ISSUE #61: Improve doc comments for public api
We are still missing:
```
  Repeater: repeater.dart:5:1
  SerialDisposable: serial_disposable.dart:3:1
  isDisposed: serial_disposable.dart:22:3
  set: serial_disposable.dart:6:3
  dispose: serial_disposable.dart:15:3
  StreamSubscriptionSerialDisposable: serial_disposable.dart:38:1
  WriteCharacteristicInfo: write_characteristic_info.dart:6:1
  WriteCharacteristicFailure: write_characteristic_info.dart:17:1
  Uuid: uuid.dart:6:1
 convertScanModeToArgs: scan_mode.dart:17:1
```

remonh87:
Fixes #61  

@werediver  please advise about repeater definition since I struggled with that one

### DIFF
--- a/lib/src/model/scan_mode.dart
+++ b/lib/src/model/scan_mode.dart
@@ -14,6 +14,7 @@ enum ScanMode {
   lowLatency,
 }
 
+/// Converts [ScanMode] to integer representation.
 int convertScanModeToArgs(ScanMode scanMode) {
   switch (scanMode) {
     case ScanMode.opportunistic:
@@ -25,6 +26,5 @@ int convertScanModeToArgs(ScanMode scanMode) {
     case ScanMode.lowLatency:
       return 2;
   }
-  assert(false);
-  return -1000;
+  throw AssertionError("Unknown scanmode provided.");
 }
--- a/lib/src/model/uuid.dart
+++ b/lib/src/model/uuid.dart
@@ -3,6 +3,7 @@ import 'dart:typed_data';
 import 'package:collection/collection.dart';
 import 'package:meta/meta.dart';
 
+/// Representation of a 16, 32 or 128-bit number used to identify BLE services and characteristics.
 class Uuid {
   final Uint8List data;
 
--- a/lib/src/rx_ext/repeater.dart
+++ b/lib/src/rx_ext/repeater.dart
@@ -2,6 +2,12 @@ import 'dart:async';
 
 import 'package:meta/meta.dart';
 
+/// [Repeater] sets an underlying stream up on the first subscription to
+/// the output [stream] and shuts it down when there are no more subscriptions.
+///
+/// The underlying stream can be manually shut down by calling [detach] and set
+/// up again by calling [attach] without terminating the existing subscriptions
+/// to the output stream.
 class Repeater<T> {
   final Stream<T> Function() _onListenEmitFrom;
   final Future<dynamic> Function() _onCancel;
--- a/lib/src/rx_ext/serial_disposable.dart
+++ b/lib/src/rx_ext/serial_disposable.dart
@@ -1,5 +1,7 @@
 import 'dart:async';
 
+/// A disposable resource whose underlying resource can be replaced by another resource
+/// causing automatic disposal of the previous underlying resource.
 class SerialDisposable<T> {
   SerialDisposable(this._dispose);
 
@@ -12,6 +14,7 @@ class SerialDisposable<T> {
     return newValue;
   }
 
+  /// Dispose underlying resource
   Future<void> dispose() async {
     _isDisposed = true;
     if (_value != null) {
@@ -19,6 +22,7 @@ class SerialDisposable<T> {
     }
   }
 
+  /// Returns whether or not the underlying resource is disposed
   bool get isDisposed => _isDisposed;
 
   final Future<void> Function(T) _dispose;
@@ -35,6 +39,7 @@ class _SerialAlreadyDisposed extends Error {
   String toString() => "An instance of $_type has already been disposed";
 }
 
+/// A [SerialDisposable] that constains an underlying stream subscription.
 class StreamSubscriptionSerialDisposable
     extends SerialDisposable<StreamSubscription> {
   StreamSubscriptionSerialDisposable()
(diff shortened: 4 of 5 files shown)


lib/src/model/scan_mode.dart
```diff
@@ -14,6 +14,7 @@ enum ScanMode {
   lowLatency,
 }
 
+/// Converts scanmode to integer representation.
```
werediver, line 17: What is "scanmode"? Either "[scanMode]" (ID reference or at lease `code` snippet), or "scan mode" (natural reference), please.

lib/src/model/scan_mode.dart
```diff
@@ -26,5 +27,5 @@ int convertScanModeToArgs(ScanMode scanMode) {
       return 2;
   }
   assert(false);
-  return -1000;
+  throw AssertionError("Unknown scanmode provided.");
```
werediver, line 30: `assert(false)` can be deleted now.

lib/src/model/uuid.dart
```diff
@@ -3,6 +3,7 @@ import 'dart:typed_data';
 import 'package:collection/collection.dart';
 import 'package:meta/meta.dart';
 
+/// Representation of a 128-bit number used to identify BLE services and characteristics
```
werediver, line 6: That is not exactly true. The standard ID is indeed a 128-bit lump, but in BLE shorter 16- and 32-bit formats are also used. The the length is checked in `Uuid.parse(·)`, but not in `Uuid(·)`, so it's possible construct an instance with arbitrary payload, actually.

lib/src/model/write_characteristic_info.dart
```diff
@@ -3,6 +3,8 @@ import 'package:flutter_reactive_ble/src/model/qualified_characteristic.dart';
 import 'package:flutter_reactive_ble/src/model/result.dart';
 import 'package:meta/meta.dart';
 
+/// Info containing the [QualifiedCharacteristic] and the result of the
+/// write characteristic operation.
```
werediver, line 7: ~"Info containing"? "Info" is way too abstract [to my taste]. I'd consider something like "a class", "a data structure".~

~But even better would be not to repeat the declaration ("A write-operation result."?).~

This is not a public class.

lib/src/model/write_characteristic_info.dart
```diff
@@ -3,6 +3,7 @@ import 'package:flutter_reactive_ble/src/model/qualified_characteristic.dart';
 import 'package:flutter_reactive_ble/src/model/result.dart';
 import 'package:meta/meta.dart';
 
+
```
werediver, line 6: Doesn't the formatter remove extra empty lines?

lib/src/model/uuid.dart
```diff
@@ -3,6 +3,7 @@ import 'dart:typed_data';
 import 'package:collection/collection.dart';
 import 'package:meta/meta.dart';
 
+/// Representation of a 16, 32 or 128-bit number used to identify BLE services and characteristics
```
werediver, line 6: As it's a sentence, would be nice to end it with a full stop.

lib/src/model/uuid.dart
```diff
@@ -63,13 +64,11 @@ class Uuid {
...
   @override
-  int get hashCode =>
-      data.fold(17, (hash, octet) => 37 * hash + octet.hashCode);
+  int get hashCode => data.fold(17, (hash, octet) => 37 * hash + octet.hashCode);
 
   @override
   bool operator ==(dynamic other) =>
-      other.runtimeType == runtimeType &&
-      const DeepCollectionEquality().equals(other.data, data);
+      other.runtimeType == runtimeType && const DeepCollectionEquality().equals(other.data, data);
```
werediver, line 71: Do these line really fit into 70 characters column?

I'd recommend to run a formatter on all the files, because if the format is not right, we'll get a penalty in the rating on pub.dev.

lib/src/rx_ext/serial_disposable.dart
```diff
@@ -1,5 +1,8 @@
 import 'dart:async';
 
+/// A disposable resource whose underlying resource can be replaced by another resource
+/// causing automatic disposal of the previous underlying resource.
+///
```
werediver, line 5: The empty line is not necessary.

lib/src/rx_ext/serial_disposable.dart
```diff
@@ -35,8 +40,7 @@ class _SerialAlreadyDisposed extends Error {
   String toString() => "An instance of $_type has already been disposed";
 }
 
-class StreamSubscriptionSerialDisposable
-    extends SerialDisposable<StreamSubscription> {
-  StreamSubscriptionSerialDisposable()
-      : super((StreamSubscription subscription) => subscription.cancel());
+/// A serialDisable that constains an underlying stream subscription
```
werediver, line 43: It's likely a typo: "serialDisable". Also, if it's an ID reference, use the corresponding syntax; if it's a natural language, use the corresponding formatting (separate words, capitalize as needed).

And full stop in the end of a sentence. Just imagine you're reading a magazine. Same formatting is required here.

lib/src/rx_ext/repeater.dart
```diff
@@ -1,7 +1,8 @@
 import 'dart:async';
 
 import 'package:meta/meta.dart';
-
+/// Streamcontroller that resumes forwarding events from the source stream when
+/// attaching streams to it.
 class Repeater<T> {
```
werediver, line 6: I'd suggest the following.

```dart
/// [Repeater] sets an underlying stream up on the first subscription to
/// the output [stream] and shuts it down when there are no more subscriptions.
/// 
/// The underlying stream can be manually shut down by calling [detach] and set
/// up again by calling [attach] without terminating the existing subscriptions
/// to the output stream.
```

lib/src/model/uuid.dart
remonh87: we do run a formatter in the CI but also locally it does not change

PR #65, opened 2020-05-20
extract connect device to separate component

remonh87:


lib/src/reactive_ble.dart
```diff
@@ -335,54 +336,15 @@ class FlutterReactiveBle {
...
-            "disconnectFromDevice", args.writeToBuffer());
-      },
-    );
-
-    return initialize()
-        .asStream()
-        .asyncExpand((_) => autoconnectingRepeater.stream);
-  }
+  }) =>
+      initialize().asStream().asyncExpand((_) => DeviceConnector(
```
remonh87, line 340: Idea is later to inject the device connector in this component

lib/src/device_connector.dart
```diff
@@ -0,0 +1,71 @@
...
+import 'package:flutter_reactive_ble/src/generated/bledata.pb.dart' as pb;
+import 'package:flutter_reactive_ble/src/model/connection_state_update.dart';
+import 'package:flutter_reactive_ble/src/rx_ext/repeater.dart';
+import 'package:meta/meta.dart';
+
+import 'model/uuid.dart';
+
+class DeviceConnector {
+  const DeviceConnector({
+    @required MethodChannel bleMethodChannel,
```
werediver, line 11: I would definitely prefer to manage all the IPC in a single place (like in the facade, even though it's questionable, whether a thing called a "facade" is the right place to take care of IPC).

Do you think it's feasible?

test/device_connector_test.dart
```diff
@@ -0,0 +1,83 @@
...
+          );
+        });
+
+        test('It emits connection updates for that device', () {
+          expect(
+              _result, emitsInOrder(<ConnectionStateUpdate>[updateForDevice]));
+        });
+
+        test('It invokes method connect device', () async {
+          await _result.first;
```
werediver, line 65: Ideally, this should be in a set-up block.

test/device_connector_test.dart
```diff
@@ -0,0 +1,83 @@
...
+          verify(_methodChannel.invokeMethod<void>('connectToDevice', any))
+              .called(1);
+        });
+
+        test('It invokes method disconnect when stream is cancelled', () async {
+          final subscription = _result.listen((event) {});
+
+          await subscription.cancel();
```
werediver, line 73: Ideally, these lines should be in a set-up block.

lib/src/plugin_controller.dart
```diff
@@ -0,0 +1,34 @@
+import 'package:flutter/services.dart';
+import 'package:flutter_reactive_ble/src/converter/args_to_protubuf_converter.dart';
+import 'package:flutter_reactive_ble/src/model/uuid.dart';
+
+class PluginController {
+  const PluginController(this.converter, this.bleMethodChannel);
```
werediver, line 6: I'd used named parameters here, because they don't seem to be obvious or easy-to-guess from the class name.

lib/src/reactive_ble.dart
```diff
@@ -98,14 +103,21 @@ class FlutterReactiveBle {
...
   /// operation is triggered.
   Future<void> initialize() async {
+    _pluginController ??=
+        PluginController(ArgsToProtobufConverter(), _methodChannel);
     _initialization ??= _methodChannel.invokeMethod("initialize");
+    _deviceConnector ??= DeviceConnector(
+      connectionStateUpdateStream: connectedDeviceStream,
+      pluginController: _pluginController,
+    );
     _prescanConnector = PrescanConnector(
```
werediver, line 121: This suspicious line... (just so that we remember)

remonh87: it is very suspicious. Next pr will to move the prescan connector inside the deviceconnector. I have a nice idea how to fix this

PR #74, opened 2020-06-16
Darft create unittests reactive ble

remonh87:
@werediver  I create a small setup for how to test the facade.  Since I do not want to sacrifice the API I needed to compromise a bit. curious about your opion, I am not super convinced. Else we can also decide to make unittests for the next major version.


remonh87:
Like discussed we go for better approach

PR #75, opened 2020-06-18
fix propagating stream error in devicescanner

### LINKED ISSUE #73: No catchable exception when BT is disabled in scanRepeater
**Describe the bug**
When BT is disabled in Android and the scan is started there is an exception internally but the exception is not handled to the calling program and therefore cannot be handled by the app.

**To Reproduce**
1. In Android switch off BT
2. start _scanner.startScan(null); from the example program
3. Check the console for failures

**Expected behavior**
We need a stream event for exceptions and some additional information about the type of exception

**Additional context**

```
I/flutter (21735): ErrorValue: Exception: GenericFailure<ScanFailure>(code: ScanFailure.unknown, message: "Bluetooth disabled (code 1)")
I/flutter (21735): #0      Result.dematerialize.<anonymous closure> (package:flutter_reactive_ble/src/model/result.dart:22:13)
I/flutter (21735): #1      Result.iif (package:flutter_reactive_ble/src/model/result.dart:34:21)
I/flutter (21735): #2      Result.dematerialize (package:flutter_reactive_ble/src/model/result.dart:15:28)
I/flutter (21735): #3      FlutterReactiveBle.scanForDevices.<anonymous closure>.<anonymous closure> (package:flutter_reactive_ble/src/reactive_ble.dart:274:49)
I/flutter (21735): #4      _MapStream._handleData (dart:async/stream_pipe.dart:229:31)
I/flutter (21735): #5      _ForwardingStreamSubscription._handleData (dart:async/stream_pipe.dart:166:13)
I/flutter (21735): #6      _rootRunUnary (dart:async/zone.dart:1192:38)
I/flutter (21735): #7      _CustomZone.runUnary (dart:async/zone.dart:1085:19)
I/flutter (21735): #8      _CustomZone.runUnaryGuarded (dart:async/zone.dart:987:7)
I/flutter (21735): #9      _BufferingStreamSubscription._sendData (dart:async/stream_impl.dart:339:11)
I/flutter (21735): #10     _BufferingStreamSubscription._add (dart:async/stream_impl.dart:266:7)
I/flutter (21735): #11     _ForwardingStreamSubscription._add (dart:async/stream_pi

```

remonh87:
Fixes #73

### DIFF
--- a/lib/src/device_scanner.dart
+++ b/lib/src/device_scanner.dart
@@ -1,9 +1,11 @@
 import 'dart:async';
 
+import 'package:flutter_reactive_ble/flutter_reactive_ble.dart';
 import 'package:flutter_reactive_ble/src/plugin_controller.dart';
 import 'package:flutter_reactive_ble/src/rx_ext/repeater.dart';
 import 'package:flutter_reactive_ble/src/rx_ext/serial_disposable.dart';
 import 'package:meta/meta.dart';
+import 'package:pedantic/pedantic.dart';
 
 import 'discovered_devices_registry.dart';
 import 'model/discovered_device.dart';
@@ -48,6 +50,8 @@ class DeviceScanner {
     final completer = Completer<void>();
     _currentScanSession =
         ScanSession(withServices: withServices, future: completer.future);
+    // Make sure completing a future with an error does not lead to an unhandled exception.
+    unawaited(completer.future.catchError((Object e, StackTrace s) {}));
 
     final scanRepeater = Repeater(
       onListenEmitFrom: () =>
@@ -55,6 +59,12 @@ class DeviceScanner {
         (Object e, StackTrace s) {
           if (!completer.isCompleted) {
             completer.completeError(e, s);
+            if (e is Exception)
+              throw e;
+            else if (e is Error)
+              throw e;
+            else
+              throw Exception(e);
           }
         },
       ),
--- a/example/pubspec.lock
+++ b/example/pubspec.lock
@@ -144,6 +144,13 @@ packages:
       url: "https://pub.dartlang.org"
     source: hosted
     version: "1.6.4"
+  pedantic:
+    dependency: transitive
+    description:
+      name: pedantic
+      url: "https://pub.dartlang.org"
+    source: hosted
+    version: "1.9.0"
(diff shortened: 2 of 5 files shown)


lib/src/device_scanner.dart
```diff
@@ -48,13 +50,16 @@ class DeviceScanner {
...
     _currentScanSession =
         ScanSession(withServices: withServices, future: completer.future);
+    // we do not want the future to throw the error but we are going to throw it in the Repeater
+    unawaited(completer.future.catchError((Object e, StackTrace s) {}));
 
     final scanRepeater = Repeater(
       onListenEmitFrom: () =>
           _scanStream.map((scan) => scan.result.dematerialize()).handleError(
-        (Object e, StackTrace s) {
+        (Exception e, StackTrace s) {
```
werediver, line 59: How can you require an `Exception` here, if the contract is "anything" (correct me, if I'm wrong)? This way you can only get a false sense of type-safety, because you can still get an `Error` or anything at all here.

I'd keep the parameter type `Object` and rethrow it somehow like: `if (e is Exception) throw e; else if (e is Error) throw e; else throw Exception(e);`

Note that a shorter `e is Exception ? e : e is Error ? e : Exception(e)` wouldn't work, because `Exception` and `Error` have no subclass-baseclass relationship (they both inherit from `Object` only), thus the type of the entire expression would be `Object` and a warning would pop-up.

Alternatively, a directive can be used to disable the corresponding diagnostic, but... I'd first looks into solutions within the bounds of the language.

remonh87: ah indeed I forgot about that good catch!

lib/src/device_scanner.dart
```diff
@@ -48,13 +50,21 @@ class DeviceScanner {
...
           _scanStream.map((scan) => scan.result.dematerialize()).handleError(
         (Object e, StackTrace s) {
           if (!completer.isCompleted) {
             completer.completeError(e, s);
+            if (e is Exception)
+              throw e;
+            else if (e is Error)
+              throw e;
+            else
+              throw Exception(e);
```
werediver, line 67: If we see this pattern pop-up more, we can extract it into a function.

Issue #25, opened 2020-02-23
Improve test coverage of reactive_ble.dart


remonh87:
Our test coverage can grow significant in case we add tests for reactive_ble.dart.

PR #77, opened 2020-06-20
Unittest reactive ble

remonh87:
@werediver  we now have 90% code coverage over the non generated classes. The  remaining 10% would be "silly" tests to check if assert != null is working and test toString methods from the model. Let's not do this :)

### DIFF
--- a/lib/src/converter/args_to_protubuf_converter.dart
+++ b/lib/src/converter/args_to_protubuf_converter.dart
@@ -10,8 +10,6 @@ class ArgsToProtobufConverter {
     Map<Uuid, List<Uuid>> servicesWithCharacteristicsToDiscover,
     Duration connectionTimeout,
   ) {
-    assert(id != null);
-
     final args = pb.ConnectToDeviceRequest()..deviceId = id;
 
     if (connectionTimeout != null) {
--- a/lib/src/reactive_ble.dart
+++ b/lib/src/reactive_ble.dart
@@ -24,6 +24,22 @@ class FlutterReactiveBle {
 
   factory FlutterReactiveBle() => _sharedInstance;
 
+  ///Create a new instance where injected depedencies are used.
+  @visibleForTesting
+  FlutterReactiveBle.witDependencies({
+    @required PluginController pluginController,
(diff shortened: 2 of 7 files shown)


lib/src/converter/args_to_protubuf_converter.dart
```diff
@@ -10,8 +10,6 @@ class ArgsToProtobufConverter {
     Map<Uuid, List<Uuid>> servicesWithCharacteristicsToDiscover,
     Duration connectionTimeout,
   ) {
-    assert(id != null);
```
remonh87, line 13: I removed this since we do not do this consistenly for other methods in this class

lib/src/reactive_ble.dart
```diff
@@ -50,7 +66,7 @@ class FlutterReactiveBle {
       Repeater(onListenEmitFrom: () async* {
         await initialize();
         yield* _deviceConnector.deviceConnectionStateUpdateStream;
-      }).stream
+      }).stream.asBroadcastStream()
```
remonh87, line 69: when testing I found out that this needs to become a broadcast stream btw I can also make it a broadcast repeater

werediver: If there is already a repeater, I'd go for a broadcast repeater. In this case it doesn't really matter, but the work differently.

lib/src/converter/args_to_protubuf_converter.dart
werediver: Sounds fine to me. If we decide we need such asserts, it would be better to scatter them over the whole library consistently.

lib/src/reactive_ble.dart
werediver: Hm. For a  repeater to work here, `_deviceConnector.deviceConnectionStateUpdateStream` should either be recreated on every call of `.deviceConnectionStateUpdateStream` getter or be a broadcast stream.

remonh87: hmm good catch indeed let's keep it as is for now this looks the most simplest solution then

remonh87:
Fixes #25

PR #82, opened 2020-07-02
Remove redundant `initialize()` call

werediver:


lib/src/reactive_ble.dart
```diff
@@ -176,8 +176,6 @@ class FlutterReactiveBle {
     QualifiedCharacteristic characteristic, {
     @required List<int> value,
   }) async {
-    await initialize();
-
     await initialize();
```
werediver, line 179: Which doesn't _really_ hurt, but not needed either.

remonh87: ow wow good catch not super efficient indeed

PR #84, opened 2020-07-07
Prefer CBAdvertisementDataLocalName over peripheral.name

### LINKED ISSUE #83: Prefer CBAdvertisementDataLocalName over peripheral.name
**Describe the bug**
We found a bug in an application using flutter_reactive_ble, this bug resulted in a device showing a different device's name. We believe the cause of the problem is described in the StackOverflow link below.  

**To Reproduce**
When using peripheral.name to define a devices name, the iOS CoreBluetooth API is not able to change the name of the device. Therefore the CBAdvertisementDataLocalName should be used. 
https://stackoverflow.com/questions/25938274/incorrect-ble-peripheral-name-with-ios

**Expected behavior**

- [X] I tried doing the same with a general BLE scanner application (e.g. [nRF Connect](https://duckduckgo.com/?q=nRF+connect)) and it exhibits the expected behavior as described above
Yes, previously we used our own flutter_blues fork, this flutter package uses the CBAdvertisementDataLocalName and the bug occured after releasing our application after implementing flutter_reactive_ble

**Smartphone / tablet**
 - OS: iOS
 - Package version: [e.g. 2.3.0]

**Peripheral device**
- Vendor, model: CUSTOM
- Does it run a custom firmware / software: yes / no

**Additional context**

danielroek:
Fixes #83

### DIFF
--- a/ios/Classes/Plugin/PluginController.swift
+++ b/ios/Classes/Plugin/PluginController.swift
@@ -2,6 +2,7 @@ import class CoreBluetooth.CBUUID
 import enum CoreBluetooth.CBManagerState
 import var CoreBluetooth.CBAdvertisementDataServiceDataKey
 import var CoreBluetooth.CBAdvertisementDataManufacturerDataKey
+import var CoreBluetooth.CBAdvertisementDataLocalNameKey
 
 final class PluginController {
 
@@ -34,13 +35,14 @@ final class PluginController {
             onDiscovery: papply(weak: self) { context, _, peripheral, advertisementData, rssi in
                 guard let sink = context.scan?.sink
                 else { assert(false); return }
-                
+
                 let serviceData = advertisementData[CBAdvertisementDataServiceDataKey] as? ServiceData ?? [:]
                 let manufacturerData = advertisementData[CBAdvertisementDataManufacturerDataKey] as? Data ?? Data();
+                let localName = advertisementData[CBAdvertisementDataLocalNameKey] as? String ?? String();
 
                 let deviceDiscoveryMessage = DeviceScanInfo.with {
                     $0.id = peripheral.identifier.uuidString
-                    $0.name = peripheral.name ?? ""
+                    $0.name = localName
                     $0.rssi = Int32(rssi)
                     $0.serviceData = serviceData
                         .map { entry in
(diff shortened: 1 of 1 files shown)


ios/Classes/Plugin/PluginController.swift
```diff
@@ -34,13 +35,14 @@ final class PluginController {
             onDiscovery: papply(weak: self) { context, _, peripheral, advertisementData, rssi in
                 guard let sink = context.scan?.sink
                 else { assert(false); return }
-                
+
                 let serviceData = advertisementData[CBAdvertisementDataServiceDataKey] as? ServiceData ?? [:]
                 let manufacturerData = advertisementData[CBAdvertisementDataManufacturerDataKey] as? Data ?? Data();
+                let localName = advertisementData[CBAdvertisementDataLocalNameKey] as? String ?? String();
```
werediver, line 41: The documentation on [CBPeripheral.name](https://developer.apple.com/documentation/corebluetooth/cbperipheral/1519029-name) states:

> A peripheral may have two different name types: one that the device advertises and another that the device publishes in its database as its Bluetooth low energy Generic Access Profile (GAP) device name. If a peripheral has both types of names, this property returns its GAP device name.

Which suggests a name may be missing in the advertisement (e.g. depending on the device mode). I would suggest to fall back to `.name` in such a case:

```swift
let name = advertisementData[CBAdvertisementDataLocalNameKey] as? String ?? peripheral.name ?? String();
```

This way the library reverses the prioritization of `CBPeripheral.name` and may show a device name different from what a user can see in the system settings, which I don't find as an unconditionally useful change, but I agree that the aggressive caching and/or insufficient cache invalidation in the iOS BLE stack does seem to be more harmful from the hardware and/or software developer standpoint.

ios/Classes/Plugin/PluginController.swift
```diff
@@ -34,13 +35,14 @@ final class PluginController {
             onDiscovery: papply(weak: self) { context, _, peripheral, advertisementData, rssi in
                 guard let sink = context.scan?.sink
                 else { assert(false); return }
-                
+
                 let serviceData = advertisementData[CBAdvertisementDataServiceDataKey] as? ServiceData ?? [:]
                 let manufacturerData = advertisementData[CBAdvertisementDataManufacturerDataKey] as? Data ?? Data();
+                let localName = advertisementData[CBAdvertisementDataLocalNameKey] as? String ?? peripheral.name ?? String();
```
werediver, line 41: I'm not sure we should still call this "local name", because the fall back is described in the iOS docs as the name registered in a GAP database. Just `name` wouldn't cause this questions.

This is a small detail, but also not difficult to correct. I can do it myself tomorrow.

PR #92, opened 2020-07-27
Consistent logging

remonh87:
Proposal for having consistent logging on iOS and Android.

- Removed Android logging
- Added a debugLogger
- Added logs for the plugincontroller (this one makes the most sense in my opinion). For the result we provide useful errors.

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -83,9 +83,8 @@ protobuf {
 
 dependencies {
     implementation "org.jetbrains.kotlin:kotlin-stdlib-jdk7:$kotlin_version"
-    implementation 'com.polidea.rxandroidble2:rxandroidble:1.10.5'
+    implementation 'com.polidea.rxandroidble2:rxandroidble:1.11.1'
     implementation 'com.google.protobuf:protobuf-lite:3.0.1'
-    implementation 'com.jakewharton.timber:timber:4.7.1'
     implementation 'io.reactivex.rxjava2:rxkotlin:2.3.0'
     implementation 'io.reactivex.rxjava2:rxandroid:2.1.1'
 
--- a/android/src/main/kotlin/com/signify/hue/flutterreactiveble/ble/DeviceConnector.kt
+++ b/android/src/main/kotlin/com/signify/hue/flutterreactiveble/ble/DeviceConnector.kt
@@ -52,13 +52,9 @@ internal class DeviceConnector(
                     ConnectionUpdateError(device.macAddress, it.message
                             ?: "Unknown error")
                 }
-                .subscribe(
-                        {
-                            updateListeners.invoke(it)
-                        },
-                        {
-                        }
-                )
+                .subscribe {
+                    updateListeners.invoke(it)
+                }
     }
 
     internal fun disconnectDevice() {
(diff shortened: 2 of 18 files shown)


android/build.gradle
```diff
@@ -83,9 +83,8 @@ protobuf {
 
 dependencies {
     implementation "org.jetbrains.kotlin:kotlin-stdlib-jdk7:$kotlin_version"
-    implementation 'com.polidea.rxandroidble2:rxandroidble:1.10.5'
+    implementation 'com.polidea.rxandroidble2:rxandroidble:1.11.1'
```
remonh87, line 86: was useful to update this one since it fixes some small issues

lib/src/debug_logger.dart
```diff
@@ -0,0 +1,23 @@
...
+  DebugLogger(void Function(Object object) printToConsole)
+      : _printToConsole = printToConsole;
+
+  void enable() {
+    _logLevel = LogLevel.verbose;
+  }
+
+  void disable() {
+    _logLevel = LogLevel.none;
+  }
```
werediver, line 13: If it's either enabled or disabled, then why to use an enumeration `_logLevel` and not boolean `enabled`?

lib/src/reactive_ble.dart
```diff
@@ -296,6 +302,9 @@ class FlutterReactiveBle {
       .clearGattCache(deviceId)
       .then((info) => info.dematerialize());
 
+  /// Subscribes to a updates from a the characteristic specified.
```
werediver, line 305: Drop `a` in `Subscribes to a updates`.

lib/src/debug_logger.dart
remonh87: I decided to put an enum because maybe in the future we would like a verbose logging and a less verboselogging, but not sure to be honest

werediver: So, be consistent: replace the enable/disable methods with `setLogLevel(·)`. Otherwise this internal flexibility cannot be leveraged without changing the public interface of this class, which would be a pity.

android/src/main/kotlin/com/signify/hue/flutterreactiveble/ble/DeviceConnector.kt
```diff
@@ -13,14 +13,13 @@ import io.reactivex.Single
...
 
 internal class DeviceConnector(
-    private val device: RxBleDevice,
-    private val connectionTimeout: Duration,
-    private val updateListeners: (update: ConnectionUpdate) -> Unit,
-    private val connectionQueue: ConnectionQueue
+        private val device: RxBleDevice,
+        private val connectionTimeout: Duration,
+        private val updateListeners: (update: ConnectionUpdate) -> Unit,
+        private val connectionQueue: ConnectionQueue
```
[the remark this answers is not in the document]
remonh87: I think detekt did this ..

lib/src/debug_logger.dart
remonh87: I agree will do that

lib/src/debug_logger.dart
```diff
@@ -0,0 +1,21 @@
...
+class DebugLogger {
+  DebugLogger(
+    String tag,
+    void Function(Object object) print,
+  )   : _tag = tag,
+        _print = print;
+
+  set logLevel(LogLevel logLevel) => _logLevel = logLevel;
+
+  void log(Object message) {
```
werediver, line 12: To be able to introduce new log-levels this method will need to be extended with the message-log-level parameter. But, I guess, that can wait.

lib/src/debug_logger.dart
```diff
@@ -0,0 +1,21 @@
...
+    if (_logLevel == LogLevel.verbose) {
+      _print('$_tag: $message');
```
werediver, line 14: I'd even consider leaving the responsibility to add `: ` (a colon followed by a space) to the caller/client (whoever creates the object). A bit less enforced constraints we have no good reasons for.

(e.g. `[BLE SCANNING] message` vs. `BLE SCANNING: message`: we don't care, but now the client have _no_ choice)

remonh87: I think that is a nice addition for the future indeed :)

PR #103, opened 2020-09-15
Add discover services

remonh87:


lib/src/plugin_controller.dart
```diff
@@ -280,10 +281,20 @@ class PluginController {
...
+            _argsToProtobufConverter
+                .createDiscoverServicesRequest(deviceId)
+                .writeToBuffer(),
+          )
+          .then(_protobufConverter.discoveredServicesFrom);
 }
 
 class PluginControllerFactory {
-  // this injection is temporarily until we moved everythin out
+  // This injection is temporarily until we moved everything out.
```
remonh87, line 297: What injection 🤔  I think it is gone.

protos/bledata.proto
```diff
@@ -114,11 +114,27 @@ message ServiceWithCharacteristics {
...
 }
 
+message DiscoverServicesRequest {
+    string deviceId = 1;
+}
+
+message DiscoverServicesInfo {
+    string deviceId = 1;
+    repeated DiscoveredService services = 2;
+    GenericFailure failure = 3;
```
remonh87, line 124: Let's remove this

- [ ] remove

lib/src/plugin_controller.dart
werediver: I deleted this comment.

protos/bledata.proto
werediver: I've pushed removal of this field in Dart part (Swift part wasn't using it at all). Kotlin is yet to be updated.

I've brushed up "test/converter/protobuf_converter_test.dart", I hope you're fine with the style I "enforced". I had a feeling that the consistency and conciseness there could be improved. Let me know if you're not happy with the changes.

werediver: Hey, @remonh87, I've deleted this field from the Android implementation in a separate commit, but I'm pretty sure I screwed it up in some way. It compiles and the tests pass, but I can't guarantee anything more than this 😅

If you don't like the approach or whatever, just strip off the latest commit or put a reverting commit on top of it and implement the change yourself.

werediver (approved):

PR #106, opened 2020-09-18
use scanrecord instead of ble devie

### LINKED ISSUE #105: Device name not updating on Android
**Describe the bug**
BLE Devices changes its name but the name in not changed on the data returned on scan. 

**To Reproduce**
Try scanning a device that changes its name on the advertising data but then on the scan results the name is not changed.

To fix this you can change ReactiveBleClient.kt:

```
                 .map { result ->
-                    ScanInfo(result.bleDevice.macAddress, result.bleDevice.name
+                    ScanInfo(result.bleDevice.macAddress, result.scanRecord.deviceName
                             ?: "",
                             result.rssi,
                             result.scanRecord.serviceData.mapKeys { it.key.uuid },
```

**Expected behavior**
Device name changes on scan results.

**Smartphone / tablet**
 - Device: Pixel 3a
 - OS: Android 10

**Peripheral device**
- Vendor, model: Custom
- Does it run a custom firmware / software: yes

**Additional context**

remonh87:
Fixes #105

### DIFF
--- a/android/src/main/kotlin/com/signify/hue/flutterreactiveble/ble/ReactiveBleClient.kt
+++ b/android/src/main/kotlin/com/signify/hue/flutterreactiveble/ble/ReactiveBleClient.kt
@@ -70,8 +70,8 @@ open class ReactiveBleClient(private val context: Context) : BleClient {
                 *filters
         )
                 .map { result ->
-                    ScanInfo(result.bleDevice.macAddress, result.bleDevice.name
-                            ?: "",
+                    ScanInfo(result.bleDevice.macAddress, result.scanRecord.deviceName
+                            ?: result.bleDevice.name ?: "",
                             result.rssi,
                             result.scanRecord.serviceData.mapKeys { it.key.uuid },
                             extractManufacturerData(result.scanRecord.manufacturerSpecificData))


android/src/main/kotlin/com/signify/hue/flutterreactiveble/ble/ReactiveBleClient.kt
```diff
@@ -70,7 +70,7 @@ open class ReactiveBleClient(private val context: Context) : BleClient {
                 *filters
         )
                 .map { result ->
-                    ScanInfo(result.bleDevice.macAddress, result.bleDevice.name
+                    ScanInfo(result.bleDevice.macAddress, result.scanRecord.deviceName
```
werediver, line 73: I specifically [mentioned](https://github.com/PhilipsHue/flutter_reactive_ble/issues/105#issuecomment-693328205) falling back to the "standard" device name to be on the safe side.

Is there a reason _not_ to do that on Android?

remonh87: whoops forgot to push

werediver (approved):
Great, now the iOS and Android implementations are in sync!

PR #107, opened 2020-09-18
Update podspec (iOS/CocoaPods configuration)

### LINKED ISSUE #104: Error using in Flutter add to app module on iOS
**Describe the bug**
Cannot use this library on a Flutter "add to app" module on iOS. There are errors on the pod install step when building the frameworks.

**To Reproduce**
- Create flutter module for add to app: 
```
flutter create --template module my_flutter
```
- On this module add flutter_reactive_ble as a dependency
- Building frameworks fails: 
```
flutter build ios-framework --output=frameworks
```

There is an error about the iOS version that can be fixed editing the module´s .ios/Podfile and setting:

```    
platform :ios, '10.0'
```

Also there is an error about Swift version not defined that can be fixed adding this line on fluttter_reactive_ble/ios/flutter_reactive_ble.podspec:

```
s.swift_version       = '4.0'
```

**Expected behavior**

**Smartphone / tablet**

**Peripheral device**

**Additional context**

werediver:
Fixes #104 (probably)

### DIFF
--- a/ios/flutter_reactive_ble.podspec
+++ b/ios/flutter_reactive_ble.podspec
@@ -1,14 +1,11 @@
-#
-# To learn more about a Podspec see http://guides.cocoapods.org/syntax/podspec.html
-#
 Pod::Spec.new do |s|
   s.name             = 'flutter_reactive_ble'
   s.version          = '0.0.1'
-  s.summary          = 'A new flutter plugin project.'
+  s.summary          = 'Bluetooth Low Energy (BLE) Flutter plug-in'
   s.description      = <<-DESC
-A new flutter plugin project.
+Bluetooth Low Energy (BLE) Flutter plug-in
                        DESC
-  s.homepage         = 'http://example.com'
+  s.homepage         = 'https://github.com/PhilipsHue/flutter_reactive_ble'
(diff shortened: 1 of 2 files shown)


ios/flutter_reactive_ble.podspec
```diff
@@ -17,11 +14,9 @@ A new flutter plugin project.
   s.dependency 'Flutter'
   s.dependency 'Protobuf', '~> 3.5'
   s.dependency 'SwiftProtobuf', '~> 1.0'
-
-
   s.ios.deployment_target = '10.0'
   s.osx.deployment_target = '10.13'
   s.pod_target_xcconfig = { 'GCC_PREPROCESSOR_DEFINITIONS' => '$(inherited) GPB_USE_PROTOBUF_FRAMEWORK_IMPORTS=1' }
-
+  s.swift_version       = '4.2'
```
werediver, line 20: This is the (technically) important part.

remonh87: does it make sense to consider other deployment targets? like 11 at least?

werediver: iOS 10.0 or newer is just what we need API-wise.

PR #117, opened 2020-10-07
Fix issue devicescan android on hot restart

### LINKED ISSUE #114: "Scanning parameters are not set" on Hot Restart
**Describe the bug**
On Android, I consistently have a failure on the first `scanForDevices` after a Hot Restart.

**To Reproduce**

This program have a single button to start a bluetooth scan
```
import 'package:flutter/material.dart';
import 'package:flutter_reactive_ble/flutter_reactive_ble.dart';

void main() => runApp(MaterialApp(
      home: Scaffold(
        body: BluetoothApp(),
      ),
    ));

final _client = FlutterReactiveBle();

class BluetoothApp extends StatelessWidget {
  @override
  Widget build(BuildContext context) {
    return SafeArea(
      child: RaisedButton(
        onPressed: () {
          _client.scanForDevices(
              withServices: [], scanMode: ScanMode.lowLatency).listen((device) {
            print('Found ${device.name}');
          }, onError: (e) {
            print('Scan error $e');
          }, cancelOnError: true);
        },
        child: Text('Start scan'),
      ),
    );
  }
}
```
- Run the program and press "Start scan"
- During the scan, execute a Hot Restart
- After the application restart, press "Start scan" again
- It fails with `Scan error Exception: GenericFailure<ScanFailure>(code: ScanFailure.unknown, message: "Scanning parameters are not set")`
- Press a second time, it works now.

**Expected behavior**
It would improve developer experience if the library was more resilient to this case.

**Smartphone / tablet**
I can consistently reproduce on this 3 devices:
Motorola G5 (Android 8.1.0), Huawei P9 lite (Android 7.0), Nokia android One (Android 9)

Package version: 2.5.0

**Peripheral device**
NA

**Additional context**
This error message has already been reported here: https://github.com/PhilipsHue/flutter_reactive_ble/issues/87 but I don't known if this is the same issue.

remonh87:
Fixes #114

### DIFF
--- a/android/src/main/kotlin/com/signify/hue/flutterreactiveble/channelhandlers/ScanDevicesHandler.kt
+++ b/android/src/main/kotlin/com/signify/hue/flutterreactiveble/channelhandlers/ScanDevicesHandler.kt
@@ -46,15 +46,19 @@ class ScanDevicesHandler(private val bleClient: com.signify.hue.flutterreactiveb
     }
 
     fun stopDeviceScan() {
-        scanForDevicesDisposable.dispose()
-        scanParameters = null
+        if (this::scanForDevicesDisposable.isInitialized) scanForDevicesDisposable.let {
+            if (!it.isDisposed) {
+                it.dispose()
+                scanParameters = null
+            }
+       }
     }
 
     fun prepareScan(scanMessage: pb.ScanForDevicesRequest) {
+        stopDeviceScan()
         val filter = scanMessage.serviceUuidsList
                 .map { ParcelUuid(UuidConverter().uuidFromByteArray(it.data.toByteArray())) }
         val scanMode = createScanMode(scanMessage.scanMode)
-
         scanParameters = ScanParameters(filter, scanMode, scanMessage.requireLocationServicesEnabled)
     }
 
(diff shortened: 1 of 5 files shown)


android/src/main/kotlin/com/signify/hue/flutterreactiveble/channelhandlers/ScanDevicesHandler.kt
```diff
@@ -46,15 +46,19 @@ class ScanDevicesHandler(private val bleClient: com.signify.hue.flutterreactiveb
     }
 
     fun stopDeviceScan() {
-        scanForDevicesDisposable.dispose()
-        scanParameters = null
+        if (this::scanForDevicesDisposable.isInitialized) scanForDevicesDisposable.let {
```
werediver, line 49: Why `scanForDevicesDisposable.let {}` is needed here? In `this::scanForDevicesDisposable.isInitialized` you've already assumed `scanForDevicesDisposable` is non-null (if that is relevant).

werediver (approved):

android/src/main/kotlin/com/signify/hue/flutterreactiveble/channelhandlers/ScanDevicesHandler.kt
remonh87: It is a bit of sugar to put it in a let block not per se needed but also no harm. Btw I did .let and not ?.let so indeed I do not check on nullability (Detekt would also complain in that case)

PR #120, opened 2020-10-08
Fix pubspec sorting

dnfield:
The sort_pub_dependencies lint is currently broken, but soon will be fixed.

This fixes the sorting if the pubspec here in preparation for that fix.

/cc @pq


remonh87 (approved):
Thank you @dnfield  for fixing this

PR #126, opened 2020-10-19
Export `DiscoveredService` class

### LINKED ISSUE #125: Class DiscoveredService isn't exported
I'm trying to create a function that discovers services for a deviceId and returns the result. It's really just a small change from the function in the example project.
I'm changing
`Future<void> discoverServices(String deviceId) async {
    ...
}`
to
`Future<List<DiscoveredService>> discoverServices(String deviceId) async {
    ...
}`

I'm having troubles importing/using the type `DiscoveredService` since the file `discovered_services.dart` isn't exported in `flutter_reactive_ble.dart` in the library.

Is this intentional or could we add `discovered_services.dart` to the list of exported files in `flutter_reactive_ble.dart`? If so I'm happy to make a PR with that small change.

werediver:
Fixes #125

### DIFF
--- a/lib/src/converter/protobuf_converter.dart
+++ b/lib/src/converter/protobuf_converter.dart
@@ -7,7 +7,7 @@ import 'package:flutter_reactive_ble/src/model/clear_gatt_cache_error.dart';
 import 'package:flutter_reactive_ble/src/model/connection_priority.dart';
 import 'package:flutter_reactive_ble/src/model/connection_state_update.dart';
 import 'package:flutter_reactive_ble/src/model/discovered_device.dart';
-import 'package:flutter_reactive_ble/src/model/discovered_services.dart';
+import 'package:flutter_reactive_ble/src/model/discovered_service.dart';
 import 'package:flutter_reactive_ble/src/model/generic_failure.dart';
 import 'package:flutter_reactive_ble/src/model/qualified_characteristic.dart';
 import 'package:flutter_reactive_ble/src/model/result.dart';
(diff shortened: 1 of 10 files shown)


remonh87 (approved):

lib/src/converter/protobuf_converter.dart
```diff
@@ -7,7 +7,7 @@ import 'package:flutter_reactive_ble/src/model/clear_gatt_cache_error.dart';
 import 'package:flutter_reactive_ble/src/model/connection_priority.dart';
 import 'package:flutter_reactive_ble/src/model/connection_state_update.dart';
 import 'package:flutter_reactive_ble/src/model/discovered_device.dart';
-import 'package:flutter_reactive_ble/src/model/discovered_services.dart';
+import 'package:flutter_reactive_ble/src/model/discovered_service.dart';
```
remonh87, line 10: just a small s maybe something to consider for us to revise to avoid confusion

werediver: Likely we had some different classes (or different naming) there before we settled on the current implementation.

lib/src/model/discovered_service.dart
```diff
@@ -22,5 +22,3 @@ class DiscoveredService extends $DiscoveredService {
   @CustomEquality(DeepCollectionEquality())
   final List<DiscoveredService> includedServices;
 }
-
-enum DiscoverServicesFailure { unknown }
```
werediver, line 26: This is also a leftover.

werediver:
Should be good now, but I was working with an older version of Flutter SDK.

PR #141, opened 2020-11-12
Lower minsdk android

### LINKED ISSUE #52: minSdk API lvl 24 as a lib requirement
**Describe the bug**
The libs minSdk requirement for Android should be lowered, since BLE has been introduced in API level 18/19, but the library is requiring API level 24

**To Reproduce**
Try to build the example app with minSdk set to API lvl 19

**Expected behavior**
Should be building while i drink my coffee

- [x] I tried doing the same with a general BLE scanner application (e.g. [nRF Connect](https://duckduckgo.com/?q=nRF+connect)) and it exhibits the expected behavior as described above

**Smartphone / tablet**
 - Device: Nexus 4, 4.4 Kitkat

**Peripheral device**
- 

**Additional context**
-

remonh87:
Fixes #52

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+## 2.7.0
+
+* Fix #137 by making sure event channels are initialized early.
+* Add support for Android 5 and 6.
+
 ## 2.6.1+1
 
 * Revert protobuf upgrade on Android.
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -46,7 +46,7 @@ android {
         }
     }
     defaultConfig {
-        minSdkVersion 24
+        minSdkVersion 21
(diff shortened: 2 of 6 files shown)


remonh87:
I do not want to go lower then android 5 since that would cause some more challenges regarding multidex etc. Also now our library covers around 95% of the total android device market which should be sufficient. By lowering it to kitkat we would gain only 1.4%

CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
+## 2.7.0
+
+* Fix #137 by making sure eventchannels are initialized early.
```
werediver, line 3: Human language, please: event channels, not eventchannels 🙂

werediver (approved):

Issue #115, opened 2020-10-06
Conflict Between Firebase Packages and Flutter_reactive_blue


plyrek:
The Firebase Packages:
  firebase_core: ^0.5.0
  firebase_auth: ^0.18.0+1
  cloud_firestore: ^0.14.0+2
recently updated to use protobuf-javalite - 3.110.jar which conflicts with the jar file protobuf-lite-3.0.1.jar used in your package making any app which uses Firebase incompatible with your package. You are not alone, several ble packages use this jar and are incompatible, Your package however, was very recently updated and seems active and I thought you would like to know.

**To Reproduce**
Add the following to pubspec.yaml and try to build an android app
firebase_core: ^0.5.0
firebase_auth: ^0.18.0+1
cloud_firestore: ^0.14.0+2
flutter_reactive_ble: ^2.5.1

**Expected behavior**
A clean build to an Android device or emulator

**Actual behavior**
Build fails with the following errors
1 exception was raised by workers:
  java.lang.RuntimeException: Duplicate class com.google.protobuf.AbstractMessageLite found in modules protobuf-javalite-3.11.0.jar (com.google.protobuf:protobuf-javalite:3.11.0) and protobuf-lite-3.0.1.jar (com.google.protobuf:protobuf-lite:3.0.1)

remonh87:
@plyrek  thanks for reporting the issue. I couldn't reproduce it by using the steps you supplied. What I did was following: 

1. Create flutter sample project 
2. Integrate the dependencies like you said.
3. Build it on a google pixel2 device (which was successful).

I did however upgrade the protobuf dependencies and will integrate them soon in master. Can you test it with your project again? Else I need a bit more info: 

- output of `flutter doctor`
- build.gradle file where you declare the depedencies. If that is problematic please let me know which gradle version, external depedencies and android tools you are using

plyrek:
Thank you so much for your quick action on this issue. The updating of protobuf dependencies did the trick! It is contributors like you that make the Flutter community such a great one!

archanpaul:
The problem persists with following : 

**pubspec.yaml** 

firebase_core: ^0.5.2
firebase_auth: ^0.18.3
cloud_firestore: ^0.14.3
firebase_remote_config: ^0.4.2
firebase_admob: ^0.10.2
flutter_reactive_ble: ^2.6.1+1

**Error** 

> 1 exception was raised by workers:
[...]

PR #143, opened 2020-11-16
Fix protobuf issues

remonh87:
Fix for  #115. Also saw that our pub.dev score was lower because we used older version of protobuf generated code

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,7 @@
 ## 2.7.1
 
 * Fix #115 by updating to new protobuf lite on Android.
-* Regenerate dart protobuf code.
+* Regenerate Dart and Swift protobuf code.
 
 ## 2.7.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+## 2.7.1
+
+* Fix #115 by updating to new protobuf lite on Android.
+* Regenerate dart protobuf code.
+
 ## 2.7.0
 
 * Fix #137 by making sure event channels are initialized early.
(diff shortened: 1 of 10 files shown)


CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
+## 2.7.1
+
+* Fix #115 by updating to new protobuf lite on Android.
+* Regenerated dart protobuf code.
```
werediver, line 4: A minor inconsistency: the first sentence is imperative (as in "fix an issue(!)", "duck and cover!"), the second is not (as in "fixed an issue", "died in vain").

In commit messages it is quite a good practice to use imperative: it's shorter and you don't need to choose between many English tenses. Because of the same reasons I like it in change logs as well (especially given that they _can_ be generated from commit messages), but whatever style you choose, let's try to use it consistently.

lib/src/generated/bledata.pbenum.dart
```diff
@@ -3,4 +3,4 @@
 //  source: bledata.proto
 //
 // @dart = 2.3
-// ignore_for_file: camel_case_types,non_constant_identifier_names,library_prefixes,unused_import,unused_shown_name,return_of_invalid_type
+// ignore_for_file: annotate_overrides,camel_case_types,unnecessary_const,non_constant_identifier_names,library_prefixes,unused_import,unused_shown_name,return_of_invalid_type,unnecessary_this,prefer_final_fields
```
werediver, line 6: So, this is all it takes? No changes in Kotlin and Swift files?

werediver (approved):

lib/src/generated/bledata.pbenum.dart
remonh87: the issue was only related to dart code (some violation of pedantic) because dart protobuf deprecated the `copywith` method for performance reasons

CHANGELOG.md
remonh87: yess will fix it

Issue #131, opened 2020-10-23
Example app crashing in Release scheme


dawidb-appdate:
**Describe the bug**
hey, when building android app using standard command 'flutter build apk --release' package is created, but when loaded to actual phone it crashes every time. I was managed to figure out that it's the very first line is crashing 'final _ble = FlutterReactiveBle();', so basically loading/initializing the library.
When running in debug on the same device it works well. 
Have you guys had similar problems? What could be causing the problem.

**To Reproduce**
1. Build release scheme with command 'flutter build apk --release'
2. Load the app to actual device 
3. Run app and see it crashing

**Smartphone / tablet**
 - Device: Mi MIX 2s
 - Android 9
 - 
**Additional context**
Android studio 4.0.2
Doctor summary (to see all details, run flutter doctor -v):
[✓] Flutter (Channel stable, 1.22.1, on Mac OS X 10.15.7 19H2, locale en-PL)
[✓] Android toolchain - develop for Android devices (Android SDK version 30.0.2)
[✓] Xcode - develop for iOS and macOS (Xcode 12.0.1)
[✓] Android Studio (version 4.0)

Log from phone:
`d.b.z.d: The exception was not handled due to missing onError handler in the subscribe() method call. Further reading: https://github.com/ReactiveX/RxJava/wiki/Error-Handling | java.lang.RuntimeException: Field status_ for c.d.a.a.c not found. Known fields are [private static final c.d.a.a.c c.d.a.a.c.e, private static volatile c.a.a.b1 c.d.a.a.c.f]
    at d.b.b0.b.a$h.a(Unknown Source:2)
    at d.b.b0.b.a$h.f(Unknown Source:2)
    at d.b.b0.d.h.b(Unknown Source:13)
    at d.b.b0.d.h.e(Unknown Source:25)
    at d.b.b0.e.e.i0$a.l(Unknown Source:47)
    at d.b.b0.e.e.i0$a.run(Unknown Source:8)
    at d.b.x.b.b$b.run(Unknown Source:2)
    at android.os.Handler.handleCallback(Handler.java:873)
    at android.os.Handler.dispatchMessage(Handler.java:99)
    at android.os.Looper.loop(Looper.java:201)
    at android.app.ActivityThread.main(ActivityThread.java:6806)
    at java.lang.reflect.Method.invoke(Native Method)
    at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:547)
[...]

PR #144, opened 2020-11-16
Add proguard rule to lib readme faq section

remonh87:
I think this will be very helpful since proguard removes a lot of our generated code now and users can run into weird errors

### DIFF
--- a/README.md
+++ b/README.md
@@ -197,3 +197,13 @@ iOS13 and higher
 
 iOS12 and lower
 * NSBluetoothPeripheralUsageDescription
+
+#### How to adjust ProGuard (Android)
+
+In case you are using ProGuard add the following snippet to your `proguard-rules.pro` file:
+
+```
+-keep class com.signify.hue.** { *; }
+```
+
+This will prevent issues like [#131](https://github.com/PhilipsHue/flutter_reactive_ble/issues/131) .


README.md
```diff
@@ -197,3 +197,7 @@ iOS13 and higher
 
 iOS12 and lower
 * NSBluetoothPeripheralUsageDescription
+
+#### How to add reactive ble lib to proguard rules (Android)
```
werediver, line 201: How to make it clear _why_ users would want / need to adjust the ProGuard rules for our library? As I understand, it's not obvious that some problems are caused by this.

If nothing better, maybe the following version could be a bit better (otherwise capitalize abbreviations, don't use cut words [e.g. lib], stick to official spelling for product names [e.g. ProGuard]):

```
#### How to adjust ProGuard rules (Android)
```

README.md
```diff
@@ -197,3 +197,7 @@ iOS13 and higher
...
+
+In case you are using proguard add the following snippet to your `proguard-rules.pro` file: `-keep class com.signify.hue.** { *; }`. This will prevent issues like #131 .
```
werediver, line 203: I suggest the followingn changes:

- fix the capitalization in "ProGuard"
- replace "snippet" with "rule"
- consider using a code block (instead of an in-line code span)
- replace `#131` with `[#131](https://github.com/PhilipsHue/flutter_reactive_ble/issues/131)` (it's not converted to a link automatically, at least I don't see it happening)

README.md
```diff
@@ -197,3 +197,13 @@ iOS13 and higher
...
+
+#### How to adjust ProGuard (Android)
+
+In case you are using ProGuard add the following snippet to your `proguard-rules.pro` file:
+
+```
+-keep class com.signify.hue.** { *; }
+```
+
+This will prevent issues like [#131](https://github.com/PhilipsHue/flutter_reactive_ble/issues/131) .
```
werediver, line 209: The space before the full-stop is not needed.

werediver (approved):

PR #163, opened 2020-12-03
add custom proguard file to reactive ble lib

### LINKED ISSUE #145: Include our own proguard rules to simplify usages of our library
It should not be needed for users to configure proguard themselves. We can make sure that the protobuf files are not removed by proguard by including a rules file in our library

For inspiration see: https://github.com/stripe/stripe-android/blob/master/stripe/proguard-rules.txt

remonh87:
Fix #145

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -48,6 +48,7 @@ android {
     defaultConfig {
         minSdkVersion 21
         targetSdkVersion 30
+        consumerProguardFiles 'proguard-rules.txt'
     }
     lintOptions {
         disable 'InvalidPackage'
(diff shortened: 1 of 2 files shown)


remonh87:
@werediver  just for the context because of issue #131 we need to add reactive ble to proguard exceptions but with this pr we already prevent proguard from stripping out code that we actually need. Tested this with our apps and it works great

werediver:
@remonh87 Does it mean after this p/r is merged the user won't need to manually add ProGuard rules for our library in their apps?

werediver (approved):

chipweinberger:
@werediver yes

Issue #174, opened 2020-12-18
Get rid of mockito


remonh87:
The problem is that Mockito needs codegen to comply to null safety. I would like to see if we can replace our mocks with stubs (also need to extract interfaces). This help us to not need to generate for our tests and we can start migrating and testing our library for null safety in Dart.

werediver:
Replacing the current mockito-dynamic approach with hand-written stubs is super tedious and would work much better if the stubs are generated. Now, mockito is going to offer such a stub-generator. What's wrong with it?

remonh87:
@werediver I did consider that option, had a discussion in the Flutter Study group about the best approach. The reasons for not choosing Mockito code generation  are:

-  It has still issues for example [bug](https://github.com/dart-lang/mockito/issues/297). Another dev mentioned it also does not work very well with the `@required` annotation.
- Mockito stil has no version that supports null-safety. We cannot migrate until they supply a version that supports it or at least update built_value dependency. 
- You can debate if codegen is the best option. Yes no manual labor but code generation in dart is slow and tedious, so maybe the dev experience will degrade for this.

I think for now I would propose for the manually created stub (that as you can see are quite simple). However we can consider later to code for codegen , however I did not see proper alternatives at the moment that support fully null-safety.

werediver:
It seems that your main argument is mockito provides a subpar null-safety support _yet_ and there is not good alternative _at the moment_ *and* you want to make the leap for null-safety _right now_, therefore we should switch to hand-written stubs.

Well, you did the job and you are willing to continue putting effort there, so on one hand go ahead. On the other hand, though, this hand-written mocks are really a low-end solution, I can tell from my experience. The biggest downsides I can highlight are:

- a lot of boilerplate code to type
- inconsistency between the hand-crafted stubs over time (you will find better ways to stub different things with time and forget some of them later, another dev will go the same path _independently_ from you, etc.)

For me that is bad enough to wait and see how mockito converges to a good-enough solution. I suspect it should happen pretty soon, because it appears to be a de-facto standard solution for testing in Dart.

Also, look at [a comment from Sander](https://github.com/PhilipsHue/flutter_reactive_ble/pull/176#discussion_r546597933) noting some recent progress with null-safety support in mockito.

@spkersten I'd suggest to continue the discussion here, in this ticket.

spkersten:
[...]

PR #176, opened 2020-12-20
Do not use mockito

remonh87:
@werediver I  moved towards stubs and extracted interfaces without compromising our test coverage.

### DIFF
--- a/lib/src/connected_device_operation.dart
+++ b/lib/src/connected_device_operation.dart
@@ -4,21 +4,21 @@ import 'package:meta/meta.dart';
 
 class ConnectedDeviceOperation {
   ConnectedDeviceOperation({
-    @required PluginController pluginController,
-  })  : assert(pluginController != null),
-        _pluginController = pluginController;
+    @required DeviceOperationController controller,
+  })  : assert(controller != null),
+        _controller = controller;
 
-  final PluginController _pluginController;
+  final DeviceOperationController _controller;
 
   Stream<CharacteristicValue> get characteristicValueStream =>
-      _pluginController.charValueUpdateStream;
+      _controller.charValueUpdateStream;
 
   Future<List<int>> readCharacteristic(QualifiedCharacteristic characteristic) {
     final specificCharacteristicValueStream = characteristicValueStream
         .where((update) => update.characteristic == characteristic)
         .map((update) => update.result.dematerialize());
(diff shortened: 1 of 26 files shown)


remonh87:
Fixes #174

test/device_connector_test.dart
```diff
@@ -335,9 +261,30 @@ void main() {
   });
 }
 
-class _PluginControllerMock extends Mock implements PluginController {}
+class _DeviceConnectionControllerStub implements DeviceConnectionController {
```
werediver, line 264: It would be better to defer the tedious work of writing stubs to a code-generator, wouldn't it? And why to develop our own, if mockito is doing this for everyone already?

What are your arguments against mockito with code-generation, actually?

lib/src/connected_device_operation.dart
```diff
@@ -2,45 +2,78 @@ import 'package:flutter_reactive_ble/flutter_reactive_ble.dart';
...
+
+  Future<int> requestMtu(String deviceId, int mtu);
+
+  Future<List<DiscoveredService>> discoverServices(String deviceId);
+
+  Future<void> requestConnectionPriority(
+      String deviceId, ConnectionPriority priority);
+}
+
+class ConnectedDeviceOperationImpl implements ConnectedDeviceOperation {
```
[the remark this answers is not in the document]
remonh87: Can you explain me what solution you would prefer (I am also not a super big fan of impl)?

test/device_connector_test.dart
remonh87: I answered it in the [issue](https://github.com/PhilipsHue/flutter_reactive_ble/issues/174) .

spkersten: https://github.com/dart-lang/mockito/blob/master/NULL_SAFETY_README.md has some alternatives to entirely manually written stubs. With version 4.1.2, both code generation and the manual calls to `super.noSuchMethod(Invocation.method(#start, [args...], return value)` seem to work (possibly with an additional cast).

werediver:
@remonh87 At least re-target this pull-request to `null-safety` (or do you have another name in mind?) branch.

remonh87:
@werediver  done will also check the CI

werediver (approved):

PR #177, opened 2020-12-23
Replace mocks with generated stubs

remonh87:


remonh87:
We do not use verify anymore in the tests and use generated mocks as preparation for null-safety. For the rest extracted basic interfaces in order to hide nitty gritty details

werediver:
We already have a branch called "null-safety". Shall this pull-request replace that branch?

remonh87:
> We already have a branch called "null-safety". Shall this pull-request replace that branch?

Yes I want to rebase null-safety branch on master again (after we merged this in) and then start the migration

werediver:
> Yes I want to rebase null-safety branch on master again (after we merged this in) and then start the migration. 

You probably want to _reset_ "null-safety" branch to master and continue with null-safety in it.

> Mocking has proper support for null-safety now and no need for the temp solution of manual stubs

This sounds very good to me.

CC @spkersten

werediver:
> We do not use verify anymore

@remonh87 Can you elaborate on this point? Why?

remonh87:
> > We do not use verify anymore
> 
> @remonh87 Can you elaborate on this point? Why?

Yes  the tests where we used this were testing implementation details of the class rather then actual outcome/state. In regard to test coverage we remain the same and have now an advantage that our tests are not tightly coupled to internal implementation.  I made even some tests better. Where we before sometimes only checked calling a specific mock, I now made sure that the SUT emits the correct stream based on preconditions.

werediver:
@remonh87 I do not really understand where in that Fowler's [article](https://martinfowler.com/articles/mocksArentStubs.html) you read that mocks are evil and force you to test implementation details.

I only see that Fowler highlights that a (simple) stub only enables _state verification_ and a mock usually enables _behavior verification_. Now, a proper unit test should test the _behavior_ (regardless of the approach) and you can do it either _indirectly_ via state verification or _directly_ via call verification. Even from the subject article I get an impression that Fowler presents mock-based testing as an approach superior to stub-based testing (not rejecting it as a viable solution, of course).

Your removal of call verification without introducing extra state verification seems to reduce the thoroughness of the existing tests, but **most importantly** you are moving away from the _direct_ (and thus most efficient) behavior verification (toward _indirect_ behavior verification via state verification).

Please care to educate me about the ideas behind your changes _or_ at least leave the call verification be.

remonh87:
@werediver: I am not against mocks, I just want to decouple my tests more of internal implementation. For example:  in the `plugin_controller_test` we check that _args_to_protobuf_converter is called with correct arguments which is in my opinion an implementation detail. 

Argument of Martin Fowler: "When you write a mockist test, you are testing the outbound calls of the SUT to ensure it talks properly to its suppliers. A classic test only cares about the final state - not how that state was derived. Mockist tests are thus more coupled to the implementation of a method. Changing the nature of calls to collaborators usually cause a mockist test to break.Coupling to the implementation also interferes with refactoring, since implementation changes are much more likely to break tests than with classic testing."
 
That said I do see value in using `verify` for the system boundaries. In our case invoking a method channel (will bring back those tests removed a bit too much in my enthusiasm).

werediver:
Makes some sense. Thank you.

I guess, we can proceed with this p/r, especially if you clean up call-verification more conservatively.

remonh87:
@werediver  thanks was a good discussion and good learning for me as well. Can you approve the pr if you agree with the approach? Then I will start with migrating to null-safety.

werediver:
@remonh87 The pros & cons of call verification can be further discussed, because it's not a question that can be answered with a simple "good" or "bad".

werediver (approved):

PR #186, opened 2021-01-06
Null safety migration

remonh87:
@werediver  the migration is not yet releasable but I want to merge this huge pr already in.

We are at the following state: 

- I did run the dart migration tool and fixed all issues the tool couldn't resolve
- I did the same for the example app
- All tests are green and no analysis issues
- The example app works

Next steps:

- We have to many nullable types in our api so I want to make sure I can get rid of them and make meaningful defaults.
-  I need to test more for the read write and subscribe so we need to extend our example app.

### DIFF
--- a/lib/src/model/connection_state_update.dart
+++ b/lib/src/model/connection_state_update.dart
@@ -13,7 +13,7 @@ class ConnectionStateUpdate extends $ConnectionStateUpdate {
   final DeviceConnectionState connectionState;
 
   /// Field `error` is null if there is no error reported.
-  final GenericFailure<ConnectionError> failure;
+  final GenericFailure<ConnectionError/*!*/> failure;
 
   const ConnectionStateUpdate({
     @required this.deviceId,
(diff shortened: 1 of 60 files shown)


remonh87:
@werediver  if you would like to review something I would suggest to review the last 5 (maybe 6 commits)

lib/src/model/connection_state_update.dart
```diff
@@ -1,14 +1,11 @@
 import 'package:flutter_reactive_ble/src/model/generic_failure.dart';
-import 'package:functional_data/functional_data.dart';
 import 'package:meta/meta.dart';
 
-part 'connection_state_update.g.dart';
 //ignore_for_file: annotate_overrides
 
 ///Status update for a specific BLE device.
 @immutable
-@FunctionalData()
```
werediver, line 10: Are you getting rid of functional_data in this p/r?

remonh87: not anymore it was a temporary workaround but we are now using latest functional data which contains the fix.

werediver (approved):

remonh87:
We need to wait for either: https://github.com/dart-lang/mockito/issues/312 or a new version of [mocktail](https://github.com/felangel/mocktail/issues)  before making the build green

remonh87:
@werediver  the Dart team fixed the issue on mockito so we are now good to go!!

werediver:
Good news! LGTM :shipit:

PR #225, opened 2021-02-13
Upgrade null safety to latest flutter stable

remonh87:
@werediver  let's ship the first version. After this I want to extend the example app with read and write and subscribe (only for null safe) which would help us in triaging bugs and improving the API's from there. 

- Part of this pr is rebase from master
-  The following commits are new: [link1](https://github.com/PhilipsHue/flutter_reactive_ble/pull/225/commits/8c5ae7149d6d2035db660a88b7cbb8b8a6cffee5) [link2](https://github.com/PhilipsHue/flutter_reactive_ble/pull/225/commits/90d463f78d7fe718bc26c58294a14fc26a24dea1)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,6 @@
 * Fix #188 `subscribeToCharacteristic` fails when characteristic config descriptor is not present.
 * Fix #195 `scanFailure` when using backgroundmode messaging Firebase.
 
-# Main releases
 ## 2.7.1
 
 * Fix #115 by updating to new protobuf lite on Android.
@@ -126,6 +125,6 @@ Other improvements:
 
 # Null safety
 
-## 1.0.0-nullsafety.0
+## 3.0.0-nullsafety.0
 
 * Initial migration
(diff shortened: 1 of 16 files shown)


CHANGELOG.md
```diff
@@ -1,3 +1,4 @@
+# Main releases
```
werediver, line 1: Why does this title appear twice in the file? I guess you wanted to move it, not copy.

Separately from that, why the null-safe version is set to 1.0.0? It looks like it's forked from a very old version of the library. I'd rather set it to 3.0.0-nullsafety.

werediver (commented):
- Please check my comment,
- Clarify what do you mean by "let's ship"
- Open a ticket for extending the example app (or let's discuss it via an other channel) with read/write/subscribe to discuss some of the details (we may try to target some standard or widespread BLE service, like time service [which may be present on a phone by default])

CHANGELOG.md
remonh87: ah yess very good one I will change this

remonh87:
@werediver  fixed the comments and also creted #226 . With ship it I mean: After merging this I will publish first prerelease version to pub.dev so users can try it out. Would be great to have some feedback from community on this impactful change.

werediver (approved):
Who doesn't like null-safety? 😄

Issue #212, opened 2021-02-02
Just works pairing


HelloBytes:
I connect to a device that has implemented Just Works pairing. 
On iOS there appears a dialog to pair with the device but on Android this is not the case. 
You can connect with Android but there are no services available. When you connect manually in the bluetooth settings of the Android device you will see the services appear.

What is the problem with Just Works pairing?

werediver:
Should Just Work.

@remonh87 Any thoughts?

remonh87:
@HelloBytes  Are you testing with a samsung device on Android? I ran into a very interesting blogpost about justworks issues on Samsung devices. https://medium.com/@martijn.van.welie/making-android-ble-work-part-4-72a0b85cb442. Maybe this work around can help you. Else I need some more logs to further triage the issue.

HelloBytes:
@remonh87 Thank you for your response and great work.

I'm testing on a pixel 2 and a Huawei tablet. 

So the work around, I have to change the code of the plugin to let it work? 

After connecting to the device I subscribe on a characteristic and I write a characteristic get the following exceptions:

Exception: GenericFailure<CharacteristicValueUpdateError>(code: CharacteristicValueUpdateError.unknown, message: "Characteristic not found with UUID

Exception: GenericFailure<WriteCharacteristicFailure>(code: WriteCharacteristicFailure.unknown, message: "Characteristic not found with UUID 000030c9-etc..")

This is because it's not paired yet. There is no state about bonding is there?

remonh87:
@HelloBytes  we do not have a notion of bonding at the moment in our library because iOS does not export this state as far as I know (right @werediver ? ) . I did check the library of polidea and on Android I really need to force ensurebonding on some devices with justworks. 

@werediver  my suggestion would be to add a flag to `connectToDevice` called `androidEnforceBondingAfterConnect`. What do you think?

werediver:
@remonh87 Do you remember how (why) the pairing works in Hue Bluetooth app? Because (I hope! 😅) it does.

I see no point in introducing a flag. What a user would do with it? Set it permanently to `true`? What's the point of having the flag then?

I would rather suggest to unify the library behavior as much as reasonably possible across the platforms. In this case it means handling pairing on android in a more automated way. The article you mentioned above ([ref](https://medium.com/@martijn.van.welie/making-android-ble-work-part-4-72a0b85cb442)) may be a good reference.

What do you think?

HelloBytes:
[...]

PR #229, opened 2021-02-20
do not discover services when bonding is in progress

remonh87:
While dicovering services can result in interfering bonding using the just works principle as stated in: https://medium.com/@martijn.van.welie/making-android-ble-work-part-4-72a0b85cb442


werediver (approved):
Looks fine. Does this resolve a particular ticket? Consider mentioning it in the description.

remonh87:
mitigation for #212

PR #230, opened 2021-02-20
do not discover services when bonding is in progress [null-safety]

remonh87:
Same as  #229 but then null-safety


werediver (approved):
I hope you cherry-picked it 🙂

PR #237, opened 2021-03-07
Add read write to example app

### LINKED ISSUE #226: Extend example App with read/write and subscribe operations
Extend example App with read/write and subscribe operations so it becomes easier to triage issues. It will be only implemented on the null-safety branch since in the future this becomes the main branch.

remonh87:
@werediver  I did use a basic implementation not architectural wonders. Maybe in the future we can move to riverpod or BLOC to have more proper state management. For now I think this helps users. I discovered already one issue in relation to demterialize in null-safety which I fixed.

https://user-images.githubusercontent.com/20230599/110250274-c43cf900-7f7a-11eb-8645-0380b0e8019f.mov

Fixes #226

### DIFF
--- a/example/lib/src/ui/device_detail/characteristic_interaction_dialog.dart
+++ b/example/lib/src/ui/device_detail/characteristic_interaction_dialog.dart
@@ -135,6 +135,10 @@ class _CharacteristicInteractionDialogState
               border: OutlineInputBorder(),
               labelText: 'Value',
             ),
+            keyboardType: TextInputType.numberWithOptions(
+              decimal: true,
+              signed: false,
+            ),
           ),
         ),
         Row(
--- a/example/lib/src/ui/device_detail/characteristic_interaction_dialog.dart
+++ b/example/lib/src/ui/device_detail/characteristic_interaction_dialog.dart
@@ -1,0 +1,157 @@
+import 'package:flutter/material.dart';
+import 'package:flutter_reactive_ble/flutter_reactive_ble.dart';
+import 'package:flutter_reactive_ble_example/src/ble/ble_device_interactor.dart';
+import 'package:provider/provider.dart';
+
+class CharacteristicInteractionDialog extends StatelessWidget {
+  const CharacteristicInteractionDialog({
+    required this.characteristic,
+    Key? key,
+  }) : super(key: key);
+  final QualifiedCharacteristic characteristic;
+
+  @override
+  Widget build(BuildContext context) {
+    return Consumer<BleDeviceInteractor>(builder: (context, interactor, _) {
+      return _CharacteristicInteractionDialog(
+        characteristic: characteristic,
+        readCharacteristic: interactor.readCharacteristic,
+        writeWithResponse: interactor.writeCharacterisiticWithResponse,
+        writeWithoutResponse: interactor.writeCharacterisiticWithoutResponse,
+      );
+    });
+  }
+}
+
+class _CharacteristicInteractionDialog extends StatefulWidget {
+  const _CharacteristicInteractionDialog({
+    required this.characteristic,
+    required this.readCharacteristic,
+    required this.writeWithResponse,
+    required this.writeWithoutResponse,
+    Key? key,
+  }) : super(key: key);
+
+  final QualifiedCharacteristic characteristic;
+  final Future<List<int>> Function(QualifiedCharacteristic characteristic)
+      readCharacteristic;
+  final Future<void> Function(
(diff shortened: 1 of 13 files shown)


example/lib/src/ui/device_detail/characteristic_interaction_dialog.dart
```diff
@@ -0,0 +1,200 @@
...
+  _CharacteristicInteractionDialogState createState() =>
+      _CharacteristicInteractionDialogState();
+}
+
+class _CharacteristicInteractionDialogState
+    extends State<_CharacteristicInteractionDialog> {
+  late String readOutput;
+  late String writeOutput;
+  late String subscribeOutput;
+  late TextEditingController textEditingController;
```
werediver, line 61: Curious, I didn't know about this new keyword `late` 👍

example/lib/src/ui/device_detail/characteristic_interaction_dialog.dart
```diff
@@ -0,0 +1,200 @@
...
+    setState(() {
+      readOutput = result.toString();
+    });
+  }
+
+  List<int> _parseInput() {
+    return textEditingController.text
+        .split(',')
+        .map(
+          (value) => int.parse(value),
```
werediver, line 91: Is it clear in UI that this is a _decimal_ number input?

lib/src/connected_device_operation.dart
```diff
@@ -45,7 +45,7 @@ class ConnectedDeviceOperationImpl implements ConnectedDeviceOperation {
   Future<List<int>> readCharacteristic(QualifiedCharacteristic characteristic) {
     final specificCharacteristicValueStream = characteristicValueStream
         .where((update) => update.characteristic == characteristic)
-        .map((update) => update.result.dematerialize());
+        .map((update) => update.result.dematerialize()!);
```
werediver, line 48: Is this `result` of type `Result<X?>`? If it's just `Result<T>`, then `.dematerialize()` should return a non-optional type, so that force-unwrapping is not needed.

Please, check this.

lib/src/model/result.dart
```diff
@@ -12,8 +12,8 @@ class Result<Value, Failure> {
         _value = null; // ignore: avoid_field_initializers_in_const_classes
 
   /// Provides the value in case of success or throws [Exception] in case of failure.
-  Value dematerialize() => iif(
-        success: (value) => value!,
+  Value? dematerialize() => iif(
```
werediver, line 15: No way this is right. The type should be non-optional.

Please revert.

werediver (changes requested):

lib/src/model/result.dart
remonh87: I know this is crap but can you explain me how we should deal with `Result<void, GenericFailure<Faliure>>` ? Because this returns an object of `Result(value: null, failure: null)` and thus crash. Maybe we should use Optional.none for this

werediver: This is sort of a language design issue. The fact that the type `void` has the only possible instance, which is `null`, is self-contradicting: you can have a non-optional `void` type, but it's going to contain `null`!

Despite being named `void`, this type in Dart (as well as in other languages, take C for example) has a structure of a "unit" type, which means a type that only has a single possible instance (hence "unit"). A type with no possible instances should be called "void" (like `Never` in Swift), it is not possible to have an instance of such type (it is still useful for functions that never return [like `fatalError(·)`] and you can tell that from their type, that's really cool).

But that's weird naming with historical reasons.

I hope you see the contradiction in Dart's "sound null-safety" here 🙂

The thing is, here we need a _proper_ "unit" type with a single possible instance, which is _not null_.

We already have such a type, see [lib/src/model/unit.dart](https://github.com/PhilipsHue/flutter_reactive_ble/blob/3e8ce1c85b111d0f89e2109c61df8c8a45829f97/lib/src/model/unit.dart), and it's used in [PluginController.clearGattCache(·)](https://github.com/PhilipsHue/flutter_reactive_ble/blob/fc94bb468766339edff2e250c771fdefc24e729b/lib/src/plugin_controller.dart#L348).

I would suggest to start using the `Unit` more in public types (in public interfaces), because with Dart null-safety it apparently becomes a necessity.

remonh87: damm I was thinking yesterday night if only we had unit 🤦

remonh87:
Thanks for the review comments I think I adressed all the points (also made the input on the ui bit better)

werediver (approved):
Now looks good 👍  Nice to see the example app improving.

PR #248, opened 2021-03-12
Null safety integration

remonh87:
need to fix conflicts (during rebase)


lib/src/model/result.dart
```diff
@@ -26,8 +25,8 @@ class Result<Value, Failure> {
 
   /// Execute specific actions on success and on failure.
   T iif<T>(
-      {@required T Function(Value value) success,
-      @required T Function(Failure failure) failure}) {
+      {required T Function(Value? value) success,
+      required T Function(Failure? failure) failure}) {
```
werediver, line 29: At least the first line should be reverted. I'm less sure about the second one.

Please, try to understand semantic of this class. Don't just put `?` and `!` until it compiles. If the class requires some more refactoring, it should be done, because this is an important infrastructural class we use a lot in the library.

lib/src/model/result.dart
```diff
@@ -53,11 +52,12 @@ class Result<Value, Failure> {
...
-  bool operator ==(dynamic other) =>
+  bool operator ==(Object other) =>
+      other is Result &&
       runtimeType == other.runtimeType &&
       _value == other._value &&
       _failure == other._failure;
 
-  final Value _value;
-  final Failure _failure;
+  final Value? _value;
```
werediver, line 61: Should not [be forced to] be optional [anymore]. If needed, the class can be specialized with an optional type (`Result<int?, String>`).

lib/src/model/scan_mode.dart
```diff
@@ -26,5 +26,4 @@ int convertScanModeToArgs(ScanMode scanMode) {
     case ScanMode.lowLatency:
       return 2;
   }
-  throw AssertionError("Unknown scanmode provided.");
```
werediver, line 29: Hooray! We don't need these shitty after-switch shenanigans!

lib/src/model/write_characteristic_info.dart
```diff
@@ -1,17 +1,33 @@
...
 import 'package:flutter_reactive_ble/src/model/qualified_characteristic.dart';
 import 'package:flutter_reactive_ble/src/model/result.dart';
+import 'package:flutter_reactive_ble/src/model/unit.dart';
 import 'package:meta/meta.dart';
 
 @immutable
 class WriteCharacteristicInfo {
   final QualifiedCharacteristic characteristic;
-  final Result<void, GenericFailure<WriteCharacteristicFailure>> result;
+  final Result<Unit, GenericFailure<WriteCharacteristicFailure>?> result;
```
werediver, line 10: You see? When you need it, you can explicitly _specialize_ `Result` with optional types. `Result` itself should not add optionality under the hood.

lib/src/plugin_controller.dart
```diff
@@ -119,10 +110,10 @@ class PluginController
...
   final Logger _debugLogger;
 
-  Stream<ConnectionStateUpdate> _connectionUpdateStream;
-  Stream<CharacteristicValue> _charValueStream;
-  Stream<ScanResult> _scanResultStream;
-  Stream<BleStatus> _bleStatusStream;
+  Stream<ConnectionStateUpdate>? _connectionUpdateStream;
+  Stream<CharacteristicValue>? _charValueStream;
+  Stream<ScanResult>? _scanResultStream;
+  Stream<BleStatus>? _bleStatusStream;
```
werediver, line 116: Can these be just `late`?

lib/src/rx_ext/serial_disposable.dart
```diff
@@ -9,25 +9,24 @@ class SerialDisposable<T> {
     if (_isDisposed) {
       throw _SerialAlreadyDisposed(runtimeType);
     }
-    if (_value != null) await _dispose(_value);
+    if (_value != null) await _dispose(_value!);
     _value = newValue;
-    return newValue;
```
werediver, line 14: Hm, weird, indeed. Doesn't make sense with the return type of `Future<void>`.

lib/src/rx_ext/serial_disposable.dart
```diff
@@ -9,25 +9,24 @@ class SerialDisposable<T> {
...
   }
 
   /// Dispose underlying resource
   Future<void> dispose() async {
     _isDisposed = true;
     if (_value != null) {
-      await _dispose(_value);
+      await _dispose(_value!);
```
werediver, line 20: It's a pity Dart doesn't "remember" a certain value is not null after an explicit check 😞

lib/src/rx_ext/serial_disposable.dart
```diff
@@ -9,25 +9,24 @@ class SerialDisposable<T> {
...
     }
   }
 
   /// Returns whether or not the underlying resource is disposed
   bool get isDisposed => _isDisposed;
 
-  final Future<void> Function(T) _dispose;
+  final Future<void>? Function(T) _dispose;
```
werediver, line 27: It's not recommended to return nulls from async functions, I believe. Can we do without optionality here?

werediver (changes requested):
I think this requires some more polishing. Feel free to reach out to me to discuss any questions you may have.

lib/src/model/write_characteristic_info.dart
remonh87: What do you exactly mean by this? the optionality is linked to the failure

werediver: I made a few comments related to `result.dart` and this one implicitly refers to them.

In particular, you have made both the `_value` and `_failure` fields in `Result` _optional_, which should not be done, because when you really want one or both of them optional, you can specialize the class with optional types, _as you are doing here_.

example/ios/Flutter/Debug.xcconfig
```diff
@@ -1,3 +1,4 @@
+#include? "Pods/Target Support Files/Pods-Runner/Pods-Runner.debug.xcconfig"
 #include "Pods/Target Support Files/Pods-Runner/Pods-Runner.debug.xcconfig"
```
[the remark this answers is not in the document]
remonh87: yes it does 🤔  will remove it and check if cocoapods will add it again

lib/src/discovered_devices_registry.dart
```diff
@@ -40,8 +40,11 @@ class DiscoveredDevicesRegistryImpl implements DiscoveredDevicesRegistry {
...
   @override
-  bool deviceIsDiscoveredRecently({String deviceId, Duration cacheValidity}) =>
+  bool deviceIsDiscoveredRecently({
+    String? deviceId,
+    Duration? cacheValidity,
+  }) =>
       discoveredDevices.containsKey(deviceId) &&
-      discoveredDevices[deviceId]
-          .isAfter(getTimestamp().subtract(cacheValidity));
+      discoveredDevices[deviceId!]!
```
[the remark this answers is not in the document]
remonh87: wow good catch what was I thinking there probably autopilot migration 🤔

lib/src/model/result.dart
```diff
@@ -13,10 +13,9 @@ class Result<Value, Failure> {
 
   /// Provides the value in case of success or throws [Exception] in case of failure.
   Value dematerialize() => iif(
-        success: (value) => value,
+        success: (value) => value!,
         failure: (failure) {
-          if (failure is Exception) {
-            // ignore: only_throw_errors
+          if (failure is Exception && failure != null) {
```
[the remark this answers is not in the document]
remonh87: Unfortunately it is. When I remove analyzer says the following: `The type 'Failure? & Exception' of the thrown expression` . I see your point though how can null be ever an Exception. Looks like a error in dart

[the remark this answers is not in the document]
werediver: Could you show me the full error message, please? In the part you provided there is no statement (like "... is wrong / ..., which I hate / ... is not my color").

lib/src/model/result.dart
remonh87: Yep I agree fixed it into: 

```dart
/// Execute specific actions on success and on failure.
  T iif<T>(
      {required T Function(Value value) success,
      required T Function(Failure failure) failure}) {
    assert(_value == null || _failure == null);

    if (_failure != null) {
      return failure(_failure!);
    } else if (_value != null) {
      return success(_value!);
    } else {
      throw Exception('Both value and failure cannot be null');
    }
  }
```

lib/src/model/result.dart
remonh87: I am not 100% about this. If we create a failure we get: 

```dart
const Result.failure(this._failure)
      : assert(_failure != null),
        _value = null; 
```
how do you want to handle this case?

lib/src/plugin_controller.dart
remonh87: we do a one time initialization later on. 

```dart
 @override
  Stream<ConnectionStateUpdate> get connectionUpdateStream =>
      _connectionUpdateStream ??= _connectedDeviceRawStream
          .map(_protobufConverter.connectionStateUpdateFrom)
          .map(
        (update) {
          _debugLogger.log(
            'Received $ConnectionStateUpdate(deviceId: ${update.deviceId}, connectionState: ${update.connectionState}, failure: ${update.failure})',
          );
          return update;
        },
      );

```

With `late` keyword we would loose this since then we would get:

```dart
 Stream<ConnectionStateUpdate> get connectionUpdateStream =>
      _connectedDeviceRawStream
          .map(_protobufConverter.connectionStateUpdateFrom)
          .map(
        (update) {
          _debugLogger.log(
            'Received $ConnectionStateUpdate(deviceId: ${update.deviceId}, connectionState: ${update.connectionState}, failure: ${update.failure})',
          );
          return update;
        },
      );
```
which I think is not what we want to achieve here

lib/src/rx_ext/serial_disposable.dart
remonh87: I think it has only problem with Generic types. Because in other cases this works properly

lib/src/rx_ext/serial_disposable.dart
remonh87: actually maybe we should use unit here

remonh87:
@werediver  see last two commits I think it looks now much better. Feel free to have another peek through the library since I think it is very useful.  Also we need to wait on mockito: 5.0.2 because we need support for build2.0 (functional data is using this)

lib/src/model/result.dart
werediver: Huh, good point 🤦‍♂️ Okay, I see now, why this is needed.

There are other ways to model sum types [that would allow us to avoid optionality here], but let's just stick with this.

werediver:
@remonh87 I'll try to make another pass tomorrow. Don't hesitate to poke me, if I get late with this.

Thanks for your effort! 💪

lib/src/model/write_characteristic_info.dart
```diff
@@ -1,17 +1,33 @@
...
+    required this.result,
   });
+
+  @override
+  String toString() =>
+      "$runtimeType(characteristic: $characteristic, result: $result)";
+
+  @override
+  int get hashCode =>
+      (((17 * 37) + characteristic.hashCode) * 37 + result.hashCode);
```
werediver, line 23: Parentheses around `17 * 37` are not needed. The outermost parentheses are also not needed.

lib/src/plugin_controller.dart
```diff
@@ -39,7 +38,7 @@ abstract class DeviceOperationController {
     QualifiedCharacteristic characteristic,
   );
 
-  Future<int> requestMtuSize(String deviceId, int mtu);
+  Future<int> requestMtuSize(String deviceId, int? mtu);
```
werediver, line 41: Suspicious optionality.

lib/src/plugin_controller.dart
werediver: Your second example doesn't use `_connectionUpdateStream` at all, so not clear, what you mean. But any way, probably not applicable.

lib/src/plugin_controller.dart
```diff
@@ -262,7 +253,7 @@ class PluginController
             _argsToProtobufConverter
                 .createWriteChacracteristicRequest(characteristic, value)
                 .writeToBuffer())
-        .then(_protobufConverter.writeCharacteristicInfoFrom);
+        .then((data) => _protobufConverter.writeCharacteristicInfoFrom(data!));
```
werediver, line 256: Can force-unwrapping here be  avoided? It's unconditional, so double nasty.

lib/src/plugin_controller.dart
```diff
@@ -279,7 +270,7 @@ class PluginController
               .createWriteChacracteristicRequest(characteristic, value)
               .writeToBuffer(),
         )
-        .then(_protobufConverter.writeCharacteristicInfoFrom);
+        .then((data) => _protobufConverter.writeCharacteristicInfoFrom(data!));
```
werediver, line 273: Can force-unwrapping here be avoided? It's unconditional, so double nasty.

lib/src/plugin_controller.dart
```diff
@@ -316,14 +307,14 @@ class PluginController
   }
 
   @override
-  Future<int> requestMtuSize(String deviceId, int mtu) async {
+  Future<int> requestMtuSize(String deviceId, int? mtu) async {
```
werediver, line 310: Suspicious optionality.

lib/src/plugin_controller.dart
```diff
@@ -316,14 +307,14 @@ class PluginController
...
     _debugLogger
         .log('Request mtu size for device: $deviceId with mtuSize: $mtu');
     return _bleMethodChannel
         .invokeMethod<List<int>>(
           "negotiateMtuSize",
           _argsToProtobufConverter
-              .createNegotiateMtuRequest(deviceId, mtu)
+              .createNegotiateMtuRequest(deviceId, mtu!)
```
werediver, line 317: Just declare it non-optional...

lib/src/plugin_controller.dart
```diff
@@ -341,11 +332,11 @@ class PluginController
               .createChangeConnectionPrioRequest(deviceId, priority)
               .writeToBuffer(),
         )
-        .then(_protobufConverter.connectionPriorityInfoFrom);
+        .then((data) => _protobufConverter.connectionPriorityInfoFrom(data!));
```
werediver, line 335: Can this bang be avoided?

lib/src/plugin_controller.dart
```diff
@@ -355,7 +346,7 @@ class PluginController
               .createClearGattCacheRequest(deviceId)
               .writeToBuffer(),
         )
-        .then(_protobufConverter.clearGattCacheResultFrom);
+        .then((data) => _protobufConverter.clearGattCacheResultFrom(data!));
```
werediver, line 349: Can force-unwrapping be avoided here?

lib/src/plugin_controller.dart
```diff
@@ -367,7 +358,7 @@ class PluginController
                 .createDiscoverServicesRequest(deviceId)
                 .writeToBuffer(),
           )
-          .then(_protobufConverter.discoveredServicesFrom);
+          .then((data) => _protobufConverter.discoveredServicesFrom(data!));
```
werediver, line 361: Can force-unwrapping be avoided here?

lib/src/reactive_ble.dart
```diff
@@ -31,11 +31,11 @@ class FlutterReactiveBle {
...
   FlutterReactiveBle.witDependencies({
-    @required BleOperationController bleOperationController,
-    @required DeviceScanner deviceScanner,
-    @required DeviceConnector deviceConnector,
-    @required ConnectedDeviceOperation connectedDeviceOperation,
-    @required Logger debugLogger,
+    required BleOperationController? bleOperationController,
+    required DeviceScanner? deviceScanner,
+    required DeviceConnector? deviceConnector,
+    required ConnectedDeviceOperation? connectedDeviceOperation,
```
werediver, line 37: Required and optional is super suspicious.

lib/src/reactive_ble.dart
```diff
@@ -70,7 +70,7 @@ class FlutterReactiveBle {
   Stream<ConnectionStateUpdate> get connectedDeviceStream =>
       Repeater(onListenEmitFrom: () async* {
         await initialize();
-        yield* _deviceConnector.deviceConnectionStateUpdateStream;
+        yield* _deviceConnector!.deviceConnectionStateUpdateStream;
```
werediver, line 73: The bang should not be needed.

lib/src/reactive_ble.dart
```diff
@@ -79,29 +79,29 @@ class FlutterReactiveBle {
   /// The updates include read responses as well as notifications.
   Stream<CharacteristicValue> get characteristicValueStream async* {
     await initialize();
-    yield* _connectedDeviceOperator.characteristicValueStream;
+    yield* _connectedDeviceOperator!.characteristicValueStream;
```
werediver, line 82: Should not be needed.

lib/src/reactive_ble.dart
```diff
@@ -79,29 +79,29 @@ class FlutterReactiveBle {
...
   }
 
-  BleOperationController _bleOperationController;
+  BleOperationController? _bleOperationController;
 
-  PluginController _pluginController;
+  PluginController? _pluginController;
```
werediver, line 87: Likely should be non-optional.

lib/src/reactive_ble.dart
```diff
@@ -79,29 +79,29 @@ class FlutterReactiveBle {
...
 
   BleStatus _status = BleStatus.unknown;
 
   Stream<BleStatus> get _statusStream =>
-      _bleOperationController.bleStatusStream;
+      _bleOperationController!.bleStatusStream;
```
werediver, line 92: 🙈

lib/src/reactive_ble.dart
```diff
@@ -79,29 +79,29 @@ class FlutterReactiveBle {
...
+  Future<void>? _initialization;
 
-  DeviceConnector _deviceConnector;
-  ConnectedDeviceOperation _connectedDeviceOperator;
-  DeviceScanner _deviceScanner;
-  Logger _debugLogger;
+  DeviceConnector? _deviceConnector;
+  ConnectedDeviceOperation? _connectedDeviceOperator;
+  DeviceScanner? _deviceScanner;
+  Logger? _debugLogger;
```
werediver, line 104: 🙈🙈🙈🙈

lib/src/reactive_ble.dart
```diff
@@ -114,16 +114,16 @@ class FlutterReactiveBle {
       print,
     );
 
-    _pluginController ??= const PluginControllerFactory().create(_debugLogger);
+    _pluginController ??= const PluginControllerFactory().create(_debugLogger!);
```
werediver, line 117: Just look at every exclamation mark in this file as if you were me, please.

werediver (commented):
I'm not convinced it's worth merging, before we're sure the optionality is annotated correctly at least where it's relatively straightforward to figure out. Marking everything as optional doesn't look like a quality improvement.

lib/src/connected_device_operation.dart
```diff
@@ -9,20 +9,20 @@ abstract class ConnectedDeviceOperation {
...
   });
 
-  Stream<List<int>> subscribeToCharacteristic(
+  Stream<List<int>?> subscribeToCharacteristic(
     QualifiedCharacteristic characteristic,
     Future<void> isDisconnected,
   );
 
-  Future<int> requestMtu(String deviceId, int mtu);
+  Future<int> requestMtu(String deviceId, int? mtu);
```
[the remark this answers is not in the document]
remonh87: that is bad  from the migration I rectified it

lib/src/connected_device_operation.dart
```diff
@@ -50,15 +49,15 @@ class ConnectedDeviceOperationImpl implements ConnectedDeviceOperation {
 
     return _controller
         .readCharacteristic(characteristic)
-        .asyncExpand((Object _) => specificCharacteristicValueStream)
+        .asyncExpand((Object? _) => specificCharacteristicValueStream)
```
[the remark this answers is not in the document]
remonh87: because of  the signature of asyncExpand `  Stream<E> asyncExpand<E>(Stream<E>? convert(T event))` . There is nothing I can think of that I can do about it

lib/src/connected_device_operation.dart
```diff
@@ -67,25 +66,25 @@ class ConnectedDeviceOperationImpl implements ConnectedDeviceOperation {
...
     final specificCharacteristicValueStream = characteristicValueStream
         .where((update) => update.characteristic == characteristic)
         .map((update) => update.result.dematerialize());
 
-    final autosubscribingRepeater = Repeater<List<int>>.broadcast(
+    final autosubscribingRepeater = Repeater<List<int>?>.broadcast(
       onListenEmitFrom: () => _controller
           .subscribeToNotifications(characteristic)
-          .asyncExpand((Object _) => specificCharacteristicValueStream),
+          .asyncExpand((Object? _) => specificCharacteristicValueStream),
```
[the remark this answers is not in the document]
remonh87: It is because of the signature of asyncexpand this is fine :)

lib/src/connected_device_operation.dart
```diff
@@ -67,25 +66,25 @@ class ConnectedDeviceOperationImpl implements ConnectedDeviceOperation {
...
-    @required List<int> value,
+    required List<int> value,
   }) async =>
       _controller
           .writeCharacteristicWithoutResponse(characteristic, value)
           .then((info) => info.result.dematerialize());
 
   @override
-  Stream<List<int>> subscribeToCharacteristic(
+  Stream<List<int>?> subscribeToCharacteristic(
```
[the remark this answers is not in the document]
remonh87: fixed

lib/src/reactive_ble.dart
remonh87: wtf is this crap. This should be a late

remonh87:
@werediver  round3 finished I think now we are getting towards a stable and high quality null safe library. Shame the migration tool did migrate so many things towards nullable type. Was more manual work then I expected

lib/src/connected_device_operation.dart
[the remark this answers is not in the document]
werediver: > ... the signature of asyncExpand [is] `Stream<E> asyncExpand<E>(Stream<E>? convert(T event))`

My question is why `T` is optional (`Object?`)? Why `_control.readCharacteristic(·)` returns `Object?`? Does it make a lot of sense?

lib/src/connected_device_operation.dart
werediver: I repeat and insist: this optionality is suspicious and should be eliminated (I checked the implementation and don't see a straightforward reason why it's needed) in this and similar cases, unless you can explain _why it makes sense_.

lib/src/connected_device_operation.dart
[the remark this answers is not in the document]
werediver: No, it's not fine, but let's discuss under an earlier comment.

werediver:
@remonh87 We had a great pair programming (sort of, any way 🙂) session. Don't bother answering my numerous comments above, just work through the notes you took and that must be enough 👍

I should still make another pass after your coming updates, but I believe we should be very close to the good shape by that time.

P.S. Code with Me IntelliJ plug-in worked pretty nicely again. I'm staring to really like it. For remote work in general, but especially for open-source activity: even with people you just _can't_ meet in person you can still have a live pair programming session, it's so cool 😎

remonh87:
> @remonh87 We had a great pair programming (sort of, any way 🙂) session. Don't bother answering my numerous comments above, just work through the notes you took and that must be enough 👍
> 
> I should still make another pass after your coming updates, but I believe we should be very close to the good shape by that time.
> 
> P.S. Code with Me IntelliJ plug-in worked pretty nicely again. I'm staring to really like it. For remote work in general, but especially for open-source activity: even with people you just _can't_ meet in person you can still have a live pair programming session, it's so cool 😎

Great session I think I fixed all the comments that we talked about. Maybe you can find somewhere time to review it for (hopefully) a last time and we can integrate it. It is much better already.

lib/src/converter/args_to_protubuf_converter.dart
```diff
@@ -40,9 +39,9 @@ abstract class ArgsToProtobufConverter {
   );
 
   pb.ScanForDevicesRequest createScanForDevicesRequest({
-    @required List<Uuid> withServices,
-    @required ScanMode scanMode,
-    @required bool requireLocationServicesEnabled,
+    required List<Uuid>? withServices,
```
werediver, line 42: Hm, did we discuss this one? Don't remember; but probably yes...

lib/src/converter/protobuf_converter.dart
```diff
@@ -87,20 +87,19 @@ class ProtobufConverterImpl implements ProtobufConverter {
       connectionState: selectFrom(
         DeviceConnectionState.values,
         index: deviceInfo.connectionState,
-        fallback: (raw) => throw _InvalidConnectionState(raw),
+        fallback: (int? raw) => throw _InvalidConnectionState(raw),
```
werediver, line 90: Dart is still unable to infer it properly? 😒

werediver (approved):
I guess, it looks fine now.

:shipit:

PR #260, opened 2021-04-15
Add serviceUuids in ScanResult

### LINKED ISSUE #258: Use advertised service UUIDs (iOS)
When scanning for devices, the `serviceData` of the device is always (see note 1) empty. 
This seems to be intended (see also issue #198) but means, to know which service are advertised, one has to explicitly scan for services on the device. 

Note 1: it seems like some service data  gets picked up, e.g. exposure notification 

    ID: EC58332D-90DA-7A3C-82F3-7F420425815B 
    Services: {fd6f: [24, 27, 255, 218, 56, 245, 11, 114, 173, 241, 138, 84, 51, 196, 27, 212, 181, 93, 196, 124]}

This can easily be tested eg. in the example app by printing service data of scan results; in ble_scanner.dart, line 30 add:

    if(device.serviceData.isNotEmpty)
      print('found device ${device.name} ${device.id} with services ${device.serviceData}');

*Alternative*
At least on iOS, it is possible to get the advertised service UUIDs which would allow to eliminate the need for a scan (unsure about the situation on Android). 

*Background*
In the context I am using flutter_reactive_ble, we do not know the exact UUID of the advertised service (i.e. cannot scan for it) but we can filter the scan results by checking parts of the UUID. 

*Solution*
Instead of using `CBAdvertisementDataServiceDataKey` use `CBAdvertisementDataServiceUUIDsKey` (see attached 
[patch.txt](https://github.com/PhilipsHue/flutter_reactive_ble/files/6303337/patch.txt)

*Result*
Instead of just getting no service data from scanning, the initial scan already yields advertised service UUIDs, e.g.

    ID: BE8CF775-DF59-D5DB-3457-6AADD9C77017
    Services: {1bb51700-58b9-1411-82fb-8dd19e64568a: [27, 181, 23, 0, 88, 185, 20, 17, 130, 251, 141, 209, 158, 100, 86, 138]}

xvrh:
Fixes #258 

Add `List<Uuid> serviceUuids` in `DiscoveredDevice`


lib/src/converter/protobuf_converter.dart
```diff
@@ -60,13 +60,16 @@ class ProtobufConverterImpl implements ProtobufConverter {
       message.serviceData.map((entry) => Uuid(entry.serviceUuid.data)),
       message.serviceData.map((entry) => Uint8List.fromList(entry.data)),
     );
+    final serviceUuids =
```
remonh87, line 63: I would like to see a unittest for this functionality

remonh87 (commented):
Thanks for the PR I think in general it looks good just would like to see a unittest for the protobuf converter but for the rest it looks fine :)

remonh87:
I also will check why codemagic is not supplying build results anymore

lib/src/converter/protobuf_converter.dart
xvrh: Done

xvrh:
Thanks @remonh87 
I added the unit test for the protobuf.

werediver (commented):
Please consider addressing the comment about `growable: false` (which is really minor, but nice to have). Otherwise looks good 👍

lib/src/converter/protobuf_converter.dart
```diff
@@ -60,13 +60,16 @@ class ProtobufConverterImpl implements ProtobufConverter {
...
+        message.serviceUuids.map((entry) => Uuid(entry.data)).toList();
```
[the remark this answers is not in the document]
xvrh: Done.

werediver (approved):
Thanks 👍

remonh87 (approved):

remonh87:
We didn't forget the pr but there some issues with the CI. I will fix this tonight  so hopefully we can integrate everything soon

PR #263, opened 2021-04-24
add codemagic yaml file

remonh87:


remonh87:
@werediver  workflow editor is not the way to go anymore they use a travis like approach. This will fix automatic CI runs on prs again

remonh87:
I decided to merge it since now we have a broken state of the repo. I did check that the build runs fine

PR #271, opened 2021-05-11
fix new Dart lints for flutter/tests

jonahwilliams:
Fixes https://github.com/flutter/flutter/issues/82285

Unfortunately it looks like some of these lints are in the generated protobuf package and would need to be updated each time the file is regenerated.

### DIFF
--- a/lib/src/generated/bledata.pb.dart
+++ b/lib/src/generated/bledata.pb.dart
@@ -3,7 +3,7 @@
 //  source: bledata.proto
 //
 // @dart = 2.12
-// ignore_for_file: annotate_overrides,camel_case_types,unnecessary_const,non_constant_identifier_names,library_prefixes,unused_import,unused_shown_name,return_of_invalid_type,unnecessary_this,prefer_final_fields
+// ignore_for_file: annotate_overrides,camel_case_types,unnecessary_const,non_constant_identifier_names,library_prefixes,unused_import,unused_shown_name,return_of_invalid_type,unnecessary_this,prefer_final_fields,prefer_constructors_over_static_methods,constant_identifier_names
 
 import 'dart:core' as $core;
 
--- a/lib/src/generated/bledata.pbjson.dart
+++ b/lib/src/generated/bledata.pbjson.dart
@@ -3,7 +3,7 @@
 //  source: bledata.proto
 //
 // @dart = 2.12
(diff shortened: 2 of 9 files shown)


lib/src/model/result.dart
```diff
@@ -16,6 +16,7 @@ class Result<Value, Failure> {
         success: (value) => value!,
         failure: (failure) {
           if (failure is Exception) {
+            // ignore: only_throw_errors
             throw failure;
```
jonahwilliams, line 20: @pq this seems like a false positive. If I try to cast to Exception I get an unnecessary cast lint

pq: I think you're right.  Would you mind opening an issue on the linter?

https://github.com/dart-lang/linter/issues/new/choose

pq: Thanks!

lib/src/generated/bledata.pbjson.dart
```diff
@@ -3,7 +3,7 @@
 //  source: bledata.proto
```
werediver, line 3: @remonh87 Why do we even have pbjson-files in the repo? I doubt we need them.

lib/src/generated/bledata.pb.dart
```diff
@@ -3,7 +3,7 @@
 //  source: bledata.proto
 //
 // @dart = 2.12
-// ignore_for_file: annotate_overrides,camel_case_types,unnecessary_const,non_constant_identifier_names,library_prefixes,unused_import,unused_shown_name,return_of_invalid_type,unnecessary_this,prefer_final_fields
+// ignore_for_file: annotate_overrides,camel_case_types,unnecessary_const,non_constant_identifier_names,library_prefixes,unused_import,unused_shown_name,return_of_invalid_type,unnecessary_this,prefer_final_fields,prefer_constructors_over_static_methods,constant_identifier_names
```
werediver, line 6:
```dart
//  Generated code. Do not modify.
```

I prefer to respect this advice. What happens if we keep the file as is? Decrease in the health-rating on pub.dev? I'd rather consider accepting that, instead of manually modifying generated files (every time).

Please, strongly consider reverting all changes to the generated files.

remonh87: it has to do with the fact that linting fails on latest dart version. I already noticed that  @jonahwilliams  created an [issue](https://github.com/dart-lang/protobuf/issues/506). I also tend to wait until the issue is fixed in protobuf and integrate it then .

jonahwilliams:
I agree, I would not land this PR as is.

remonh87:
@jonahwilliams  I cherry-picked your changes because also noticed we had unused protobuf files and analysis_options.yaml was on the wrong place in the project. Issues are fixed in #280

PR #275, opened 2021-05-18
Remove "unnecessary" imports.

srawlins:
In each library where an import is removed, the library uses some elements
provided by the import, BUT there is another import which provides all of the
same elements, and at least one more which the library uses.

In this change, we remove the imports which can be simply removed in favor of
the other already present imports.

See https://github.com/dart-lang/sdk/issues/44569 for more information.

### DIFF
--- a/test/connected_device_operation_test.dart
+++ b/test/connected_device_operation_test.dart
@@ -2,6 +2,7 @@ import 'dart:async';
 
 import 'package:flutter_reactive_ble/flutter_reactive_ble.dart';
 import 'package:flutter_reactive_ble/src/connected_device_operation.dart';
+import 'package:flutter_reactive_ble/src/model/result.dart';
 import 'package:flutter_reactive_ble/src/model/uuid.dart';
 import 'package:flutter_reactive_ble/src/model/write_characteristic_info.dart';
 import 'package:flutter_reactive_ble/src/plugin_controller.dart';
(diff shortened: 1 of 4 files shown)


test/connected_device_operation_test.dart
```diff
@@ -2,10 +2,7 @@ import 'dart:async';
 
 import 'package:flutter_reactive_ble/flutter_reactive_ble.dart';
 import 'package:flutter_reactive_ble/src/connected_device_operation.dart';
-import 'package:flutter_reactive_ble/src/model/characteristic_value.dart';
-import 'package:flutter_reactive_ble/src/model/qualified_characteristic.dart';
 import 'package:flutter_reactive_ble/src/model/result.dart';
-import 'package:flutter_reactive_ble/src/model/unit.dart';
```
werediver, line 8: Prefer to import individual files instead of the library umbrella "header" (`import 'package:flutter_reactive_ble/flutter_reactive_ble.dart';`).

Please keep the individual file imports, remove the umbrella file import, and add any missing individual file imports necessary.

This relates to every files touched in this pull request.

werediver (changes requested):

test/connected_device_operation_test.dart
srawlins: Thanks for the quick review; done.

werediver (approved):
Thank you! I'm a bit surprised this hasn't been noticed and fixed before.

werediver:
Note to myself and @remonh87, whoever is going to merge this pull-request: SQUASH! (the history isn't very entertaining here)

Issue #277, opened 2021-05-20
Cannot connect to BLE device when using Firebase Messaging on Android and setting a background message handler


soyangel:
**Describe the bug**
Cannot connect to BLE device when using Firebase Messaging on Android and setting a background message handler.

**To Reproduce**
1. Configure Firebase Messaging
2. Use the code pasted below (with it cannot connect to device)
3. When commenting the line with `FirebaseMessaging.onBackgroundMessage` it connects successfully

**Expected behavior**
Should connect successfully to BLE device

**Smartphone / tablet**
 - Device: Pixel 3a
 - OS: Android 11
 - Package version: 3.1.0

**Peripheral device**
- Vendor, model: CUSTOM
- Does it run a custom firmware / software: yes 

**Additional context**
pubspec.yaml
```yaml
dependencies:
  flutter:
    sdk: flutter
  cupertino_icons: ^1.0.2
  flutter_reactive_ble: ^3.1.0
  permission_handler: ^7.1.0
  firebase_core: "^1.2.0"
  firebase_messaging: "^10.0.0"
```

main.dart
```dart 
import 'package:firebase_core/firebase_core.dart';
import 'package:firebase_messaging/firebase_messaging.dart';
import 'package:flutter/material.dart';
import 'package:flutter_reactive_ble/flutter_reactive_ble.dart';
import 'package:permission_handler/permission_handler.dart';

final flutterReactiveBle = FlutterReactiveBle();
FirebaseMessaging messaging = FirebaseMessaging.instance;

void main() {
  WidgetsFlutterBinding.ensureInitialized();

  runApp(MyApp());

  Permission.location.request();
  _initFirebase();
}

Future<void> _firebaseMessagingBackgroundHandler(RemoteMessage message) async {
  print("Handling a background message: ${message.messageId}");
}

void _initFirebase() async {
  await Firebase.initializeApp();
  FirebaseMessaging.onBackgroundMessage(_firebaseMessagingBackgroundHandler);
}

class MyApp extends StatelessWidget {
  @override
  Widget build(BuildContext context) {
    return MaterialApp(
      title: 'Flutter Demo',
      theme: ThemeData(
        primarySwatch: Colors.blue,
      ),
      home: MyHomePage(),
    );
  }
}

class MyHomePage extends StatefulWidget {
  MyHomePage({Key? key}) : super(key: key);

  @override
  _MyHomePageState createState() => _MyHomePageState();
}

[...]

PR #279, opened 2021-05-21
Fix issues fb connectivity in bg

remonh87:
Fix for #277

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -3,17 +3,16 @@ version '1.0-SNAPSHOT'
 
 buildscript {
     ext.detekt_version = '1.14.1'
-    ext.kotlin_version = '1.4.21'
+    ext.kotlin_version = '1.5.0-RC'
     repositories {
         google()
-        jcenter()
         maven {
             url "https://plugins.gradle.org/m2/"
         }
     }
 
     dependencies {
-        classpath 'com.android.tools.build:gradle:4.1.2'
+        classpath 'com.android.tools.build:gradle:4.2.1'
         classpath 'com.google.protobuf:protobuf-gradle-plugin:0.8.14'
         classpath "org.jetbrains.kotlin:kotlin-gradle-plugin:$kotlin_version"
         classpath "io.gitlab.arturbosch.detekt:detekt-gradle-plugin:$detekt_version"
@@ -24,7 +23,6 @@ buildscript {
 rootProject.allprojects {
     repositories {
         google()
-        jcenter()
     }
 }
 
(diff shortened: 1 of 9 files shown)


android/src/main/kotlin/com/signify/hue/flutterreactiveble/ble/ReactiveBleClient.kt
```diff
@@ -55,46 +60,60 @@ open class ReactiveBleClient(private val context: Context) : BleClient {
     recommended way to call varargs in java https://kotlinlang.org/docs/reference/java-interop.html#java-varargs
     */
     @Suppress("SpreadOperator")
-    override fun scanForDevices(services: List<ParcelUuid>, scanMode: ScanMode, requireLocationServicesEnabled: Boolean): Observable<ScanInfo> {
```
remonh87, line 58: after here all formatting (done by detekt)

android/build.gradle
```diff
@@ -3,17 +3,16 @@ version '1.0-SNAPSHOT'
 
 buildscript {
     ext.detekt_version = '1.14.1'
-    ext.kotlin_version = '1.4.21'
+    ext.kotlin_version = '1.5.0-RC'
```
werediver, line 6: Assuming "RC" is for "release candidate", is it okay to specify a non-stable version here? (considering only the numbered non-suffixed releases stable)

android/src/main/kotlin/com/signify/hue/flutterreactiveble/ble/ReactiveBleClient.kt
```diff
@@ -37,15 +37,20 @@ open class ReactiveBleClient(private val context: Context) : BleClient {
     private val connectionQueue = ConnectionQueue()
     private val allConnections = CompositeDisposable()
 
-    override val connectionUpdateSubject: BehaviorSubject<ConnectionUpdate> = BehaviorSubject.create()
-
     companion object {
+        // this needs to be in companion update since backgroundisolates respawn the eventchannels
+        // Fix for https://github.com/PhilipsHue/flutter_reactive_ble/issues/277
+        private val connectionUpdateBehaviorSubject: BehaviorSubject<ConnectionUpdate> =
+            BehaviorSubject.create()
```
werediver, line 44: There are multiple approaches to support multiple Flutter engines. Doing that is definitely a very good idea, but is this the only change needed?

werediver (approved):

android/build.gradle
remonh87: I got a warning that I need to move to 1.5.0 but I think I will move back to 1.4.21 indeed

android/src/main/kotlin/com/signify/hue/flutterreactiveble/ble/ReactiveBleClient.kt
remonh87: > There are multiple approaches to support multiple Flutter engines. Doing that is definitely a very good idea, but is this the only change needed?

Yes I did check the rest of the operations and this is the only one that is needed since they are either linked to the same connection subject or they will be re initialized due to the fact we use embedding v2

PR #282, opened 2021-05-26
move disposable into companion object so they survive isolate creation

remonh87:
not everything was fixed :(

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Main releases
 
+# 3.1.1+1
+
+* Fix broken read/write subscribe operations in background mode.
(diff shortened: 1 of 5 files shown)


werediver:
It did look a bit too small last time. Well, good job debugging it 👍

werediver (approved):
I don't really like this approach (as I mentioned before, it's not the only possible way to shape the solution), but as long as it works.

remonh87:
I really dislike this as well I think we really need refactor the android side a bit more to have a nice pipeline. IMO when a new isolate is spawned we should maybe reinitialize the ble on a nicer manner instead of making it survive all destructions

PR #296, opened 2021-06-22
Requesting pull request discussion for addition of raw scan record

LEMUSADR000:
Is there any way to create an accessor for raw scan responses? The manufacturer advertisement data does not contain all of the data broadcasted by the peripheral if the peripheral doesn't correctly adhere to the generally accepted Bluetooth specifications. This is only an issue on Android and is described further in this relevant [issue](https://github.com/pauldemarco/flutter_blue/issues/102#issuecomment-428087684) in the flutter_blue library.  
  
My workaround for this on the native application which my team previously supported (using Java+Kotlin) was to access the raw signed scan record bytes. This isn't something which is available just yet on this library. 

Is there any way we could think about including it? Thank you guys so much for creating this great package. I've attached a first stab attempt at implementing this raw scan record.


LEMUSADR000:
Closing and going to re-open when I pull changes from latest major release.

PR #315, opened 2021-07-25
assume connecting state on iOS

### LINKED ISSUE #183: Events different connectToAdvertisingDevice iOS vs Android
**Describe the bug**
Events on connectToAdvertisingDevice are different between iPS and Android. When connecting to peripheral device i receive the DeviceConnectionState.connecting event on Android, but not on iOS.

**To Reproduce**
1. Connect to one peripheral using connectToAdvertisingDevice()

2. Observe the DeviceConnectionState.connecting is not received on iOS.

**Expected behavior**
The DeviceConnectionState.connecting event should also received on iOS.

**Smartphone / tablet**
 - Devices: [iPhone SE , galaxy S7]
 - OS: [iOS 14.2 Android 8.0.0]
 - Package version: [2.7.1]

**Peripheral device**
- Vendor, model: pyboard rev d
- Does it run a custom firmware / software: yes

**Additional context**
The app reproducing the problem can be found at https://github.com/wolfc01/flutter_reactive_ble_uart_example  @hash d2612b3

remonh87:
same behavior as Android

fix #183 

@werediver  I started learning Swift recently so not sure if this is what you also had in mind

### DIFF
--- a/ios/Classes/Plugin/PluginController.swift
+++ b/ios/Classes/Plugin/PluginController.swift
@@ -208,7 +208,17 @@ final class PluginController {
         let timeout = args.timeoutInMs > 0 ? TimeInterval(args.timeoutInMs) / 1000 : nil
 
         completion(.success(nil))
-
+        
+        if let sink = connectedDeviceSink {
+            let message = DeviceInfo.with {
+                $0.id = args.deviceID
+                $0.connectionState = encode(.connecting)
+            }
+            sink.add(.success(message))
+        } else {
+            print("Warning! No event channel set up to report a connection update")
+        }
+        
         do {
             try central.connect(
                 to: deviceID,


ios/Classes/Plugin/PluginController.swift
```diff
@@ -221,6 +221,12 @@ final class PluginController {
                 print("Warning! No event channel set up to report a connection failure: \(error)")
                 return
             }
+            
+            sink.add(.success(DeviceInfo.with {
+                $0.id = args.deviceID
+                $0.connectionState = encode(.connecting)
+            }
+            ))
```
werediver, line 229: This is not the right place. Here we know an error has already happened while we were trying to connect.

`.connecting` should before or right after the call to `central.connect` (depends on whether we want the errors to be reported after `.connecting` or without it).

Are there other similar cases around (maybe scanning)? Do you know how it's happening on Android?

---

A separate issue is formatting, it looks off, at least in the end (the closing braces). Look how it's formatted somewhere else, please.

werediver (changes requested):
Thanks for looking into it. You're close 🙂

ios/Classes/Plugin/PluginController.swift
remonh87: @werediver  I agree with that the place is quite late. My rationale of putting it there was that  from here we have a sink that is guaranteed to be there (see the guard on the line above). However I can move that part of the code further up. What do you think? 

Other question is there a standard formatter in XCode that you use?

remonh87: Thanks for the help. I think I put it at the correct place now

ios/Classes/Plugin/PluginController.swift
```diff
@@ -208,6 +208,19 @@ final class PluginController {
         let timeout = args.timeoutInMs > 0 ? TimeInterval(args.timeoutInMs) / 1000 : nil
 
         completion(.success(nil))
+        
+        guard let sink = connectedDeviceSink
+        else {
+            print("Warning! No event channel set up to report a connection update")
+            return
+        }
```
werediver, line 216: With your current implementation this is not "just" a warning. When you're returning from here you're canceling the connection request.

One way around this is to rearrange the code in the following manner:

```swift
if let sink = connectedDeviceSink {
    let message = DeviceInfo.with {
        $0.id = args.deviceID
        $0.connectionState = encode(.connecting)
    }

    sink.add(.success(message))
} else {
    print("Warning! No event channel set up to report a connection update")
}
```

werediver (changes requested):
You're close 🙂

ios/Classes/Plugin/PluginController.swift
remonh87: @werediver  thanks for the lesson I added a new fix with your suggestion but the problem I have there is that I need to keep put back the guard method in the do block since sink is then out of scope of that block.

I considered an alternative to make the variable sink optional but that goes a bit against the purpose since `connectedDeviceSink` is already optional

werediver: At least this way the old behavior is preserved as much as possible. I'm not implying it cannot or shouldn't be changed, but that should be though through.

I hope we're reliably getting the sink here 🙄

werediver (approved):
Looks good to me 👍 Let's see how it works for people and if any other amendments are needed.

PR #335, opened 2021-08-09
Merged master into federated-plugin-structure

juliansteenbakker:
Please verify if the merge conflicts are resolved correctly. 
I had to regenerate protobuf but it generated some files that where not in the master repository like bledata.pbenum.dart, bledata.pbjson.dart and bledata.pbserver.dart. @remonh87 Do need to add these files to this PR or should they be kept out of git?


remonh87:
I removed these files since we do not use them anymore

juliansteenbakker:
> I removed these files since we do not use them anymore

Then this PR is ready to merge!

remonh87:
thanks for the help I did test the example app a bit and everything works. I will spend some time coming days to fix up the structure.

Issue #246, opened 2021-03-11
Add Linux desktop support


robert-ancell:
Flutter desktop support is now in beta, and it would be great if flutter_reactive_ble could work on both mobile and desktop. I've developed a [bluez package](https://pub.dev/packages/bluez) that interfaces with the Linux BlueZ Bluetooth stack, and I'd like to propose we add support for that.

robert-ancell:
I opened this issue to split out from https://github.com/PhilipsHue/flutter_reactive_ble/issues/112 as the Windows/MacOS solutions are likely to be quite different.

remonh87:
@robert-ancell  it took a while but I migrated the federated plugin structure. Is there any chance you can help out with the linux support for our plugin and modify the pr?

robert-ancell:
Thanks @remonh87 - I'll have a go at updating this branch to the new format.

PR #245, opened 2021-03-11
Implement Linux support

robert-ancell:
This is a branch I'm working on to add Linux Bluetooth support. I've based it on the null-safety branch which I hope will land in master soon. It's still in an early stage, but I though I'd raise it here to get feedback.

I thought this would require the [plugin to be federated](https://github.com/PhilipsHue/flutter_reactive_ble/issues/128), but it seems we can add support in a similar fashion under the current structure.

PR #347, opened 2021-08-25
Implement Linux support

robert-ancell:
https://github.com/PhilipsHue/flutter_reactive_ble/issues/246


robert-ancell:
This branch isn't 100% there, but probably a good time for a review! @remonh87 

Some issues I encountered:
- In the example program, if you discover new devices with BlueZ, then try and connect to them they are no longer available. It seems the BlueZ technique is to enable discovery and then connect during the discovery phase. Paired devices are always available.
- We can't currently subscribe to notifications, this requires more work in dbus.dart.

robert-ancell:
This builds on the old PR https://github.com/PhilipsHue/flutter_reactive_ble/pull/245 pre-federation.

packages/flutter_reactive_ble/lib/src/reactive_ble.dart
```diff
@@ -104,8 +105,13 @@ class FlutterReactiveBle {
         print,
       );
 
-      ReactiveBlePlatform.instance =
-          const ReactiveBleMobilePlatformFactory().create();
+      if (Platform.isLinux) {
```
remonh87, line 108: I am actually thinking to inject a reactiveble platform (with the mobile as default). That way mobile users are not confronted with linux specific dependencies and the other way around.

remonh87 (changes requested):
@robert-ancell  Thanks for this PR!!  I really like the fact that the native platform support is in a separate plugin because we lack the knowledge about the linux BLE stack. That said it made me think  about some things:

- How are we going to do the  maintenance and release of the plugin? I can see two approaches: 
      -  we go for this approach and make you contributor to this repo and make it official part of the repo or..
      - The advantage of the federated structure is that you can keep full control of the linux plugin and we treat it as an [non endorsed plugin](https://flutter.dev/docs/development/packages-and-plugins/developing-packages#non-endorsed-federated-plugin). Which we will fully promote and add our example app and docs.

Let me know what you prefer.

I set the pr to request changes because:

- [ ]  it is not finished (see previous comment)
- [ ]  Need to fix CI
- [ ]  Add the BSD license to the package
- [ ]  Add some docs about how to run it on linux and reference to the BlueZ plugin.

robert-ancell:
Our preference is option 1 - keep the plugin in the flutter_reactive_ble repository, thanks!

packages/flutter_reactive_ble/lib/src/reactive_ble.dart
robert-ancell: I don't know how to do this - can you point at an example / some info on what this requires?

robert-ancell:
Now builds in CI, licensing updated.

Outstanding:
- Currently doesn't have any tests.
- No specific Linux documentation added - I'm not sure what to write for these? You shouldn't need any specific actions to make this work, it will just work if BlueZ is present on the system.

remonh87:
> Now builds in CI, licensing updated.
> 
> Outstanding:
> 
> * Currently doesn't have any tests.
> * No specific Linux documentation added - I'm not sure what to write for these? You shouldn't need any specific actions to make this work, it will just work if BlueZ is present on the system.

Great progress awesome! About the docs: I was thinking if there were some specific instructions needed to run it besides starting the app on your linux machine.  If there aren't then no need to add it :)

packages/flutter_reactive_ble/lib/src/reactive_ble.dart
remonh87: I did look to other plugins and the way it is currently done by you and in the library is the most common way. Let's stick with this.

MoralCode:
I'm able to run the example app provided, however i seem to be having issues running it in my own application that already uses flutter_reactive_ble. I seem to be seeing the errors:

-  `[ERROR:flutter/runtime/dart_vm_initializer.cc(41)] Unhandled Exception: MissingPluginException(No implementation found for method initialize on channel flutter_reactive_ble_method)`
- `[ERROR:flutter/runtime/dart_vm_initializer.cc(41)] Unhandled Exception: MissingPluginException(No implementation found for method c on channel flutter_reactive_ble_method)`

 when starting to scan for devices. `Bluez` appears to be installed on my system and i have made sure that my `linux` platform directory matches the one from the example (most changes seem to be comments though). Is there something i'm missing?

jwinarske:
> * ``
> 
> 
> when starting to scan for devices. `Bluez` appears to be installed on my system and i have made sure that my `linux` platform directory matches the one from the example (most changes seem to be comments though). Is there something i'm missing?
@MoralCode It works fine for me on desktop and device.  Does it work in debug, or issue is limted to AOT?

@remonh87 is this going to get merged?  What's the holdup?

MoralCode:
> It works fine for me on desktop and device. Does it work in debug, or issue is limted to AOT?

I was most likely running it in debug mode when i posted my previous comments. I dont remember there being a difference between running via hot reload (which i assume is what you mean by AOT) vs doing a full build and run. I'm in the process of polishing the code up and making it open source, so I'll likely be working on a branch with my actual code sometime soon.

PR #392, opened 2021-09-30
Expose characteristic properties

### LINKED ISSUE #181: Expose characteristic properties (read/write/notify/...)
**Is your feature request related to a problem?**

This is a follow-up on #179.

The library should expose BLE characteristic properties ([iOS doc](https://developer.apple.com/documentation/corebluetooth/cbcharacteristic/1519010-properties)).

**Describe alternatives you've considered**

Without exposing BLE characteristic properties, the only way to judge whether a particular property is set is to attempt performing the corresponding operation and judge by the outcome (which is also problematic, because of obscure exceptions thrown by the library).

jalpedersen:
Expose characteristic properties when discovering services.
This closes #181


example/android/app/src/main/AndroidManifest.xml
```diff
@@ -6,6 +6,20 @@
     <uses-permission android:name="android.permission.BLUETOOTH" />
     <uses-permission android:name="android.permission.BLUETOOTH_ADMIN" />
 
+      <!-- Needed only if your app looks for Bluetooth devices.
+        You must add an attribute to this permission, or declare the
+        ACCESS_FINE_LOCATION permission, depending on the results when you
+      check location usage in your app. -->
+    <uses-permission android:name="android.permission.BLUETOOTH_SCAN" />
```
remonh87, line 13: this is only needed for Android12 and currently the underlying repo doesn't't support the new permissions. Can you remove these?

example/lib/src/ui/device_detail/device_interaction_tab.dart
```diff
@@ -172,15 +172,39 @@ class _ServiceDiscoveryListState extends State<_ServiceDiscoveryList> {
     super.initState();
   }
 
-  Widget _characteristicTile(QualifiedCharacteristic characteristic) =>
+  String _charactisticsSummary(DiscoveredCharacteristic c) {
+    final props = <String>[];
+    if (c.isReadable) {
+      props.add("r");
```
remonh87, line 178: let's add a bit more text since not everyone understands what `r` or `wo` means

example/pubspec.lock
```diff
@@ -399,16 +399,16 @@ packages:
   reactive_ble_mobile:
     dependency: transitive
     description:
-      name: reactive_ble_mobile
-      url: "https://pub.dartlang.org"
-    source: hosted
+      path: "../packages/reactive_ble_mobile"
```
remonh87, line 402: this is needed for development but should not be committed

packages/reactive_ble_mobile/lib/src/generated/bledata.pb.dart
```diff
@@ -10,26 +10,12 @@ import 'dart:core' as $core;
 import 'package:protobuf/protobuf.dart' as $pb;
 
 class ScanForDevicesRequest extends $pb.GeneratedMessage {
-  static final $pb.BuilderInfo _i = $pb.BuilderInfo(
```
remonh87, line 13: can you run `melos format` for the whole repo? I think this will drastically change the diff of this PR

packages/reactive_ble_mobile/lib/src/generated/bledata.pbenum.dart
```diff
@@ -0,0 +1,7 @@
+///
```
remonh87, line 1: This file is generated but we do not need it so it can be removed.

- [ ] remove

packages/reactive_ble_mobile/lib/src/generated/bledata.pbserver.dart
```diff
@@ -0,0 +1,9 @@
+///
```
remonh87, line 1: This file is generated but we do not need it so it can be removed.

- [ ] remove

packages/reactive_ble_mobile/lib/src/generated/bledata.pbjson.dart
```diff
@@ -0,0 +1,326 @@
+///
```
remonh87, line 1: This file is generated but we do not need it so it can be removed.

- [ ] Remove

packages/reactive_ble_mobile/pubspec.lock
```diff
@@ -350,9 +350,9 @@ packages:
   reactive_ble_platform_interface:
     dependency: "direct main"
     description:
-      name: reactive_ble_platform_interface
-      url: "https://pub.dartlang.org"
-    source: hosted
+      path: "../reactive_ble_platform_interface"
```
remonh87, line 353: same here the relative path is needed for development but should not be integrated

remonh87 (changes requested):
Thanks for the pr I left some small comments that need attention but in general this looks good! 👍

packages/flutter_reactive_ble/pubspec.lock
```diff
@@ -371,16 +371,16 @@ packages:
   reactive_ble_mobile:
     dependency: "direct main"
     description:
-      name: reactive_ble_mobile
-      url: "https://pub.dartlang.org"
-    source: hosted
+      path: "../reactive_ble_mobile"
```
remonh87, line 374: can you revert this one and the package below?

remonh87 (approved):
Can you only fix the pubspec.lock file but for the rest it looks good! 👍 thanks!

packages/flutter_reactive_ble/pubspec.lock
jalpedersen: All pubspec.lock files are now reverted :-)

Issue #243, opened 2021-03-10
Mac Big Sur support?


realrk69:
Now that Apple silicon enabled Macs can run iOS applications, is it possible to use this plugin to target Mac OS (Big Sur) using Flutter? Are the bindings the same? It would be great if we did not have to compile to each OS separately and it can be done on the same platform.

Thanks in advance

ekuleshov:
@realrk95 this largely depend on the Flutter's own platform support for Mac. Currently Flutter doesn't seem to support building iOS apps for desktop OSX and you have to have separate plugin for OSX.

See https://github.com/flutter/flutter/issues/33860

Issue #112, opened 2020-10-04
Support for new Desktop platforms.


JoeGilkey:
**Is your feature request related to a problem? (please describe)**
It would be very nice to have support for the new desktop platforms that Flutter has added.

**Describe the solution you'd like**
I would like to share code across an Android/iOS solution with a desktop solution. Each of the desktop platforms have ways to use BLE. For the Windows platform, it appears there is a feature request in https://github.com/timsneath/win32 to add bluetooth support.

**Describe alternatives you've considered**
I haven't found any alternatives yet for Flutter. Xamarin does have a cross platform library (https://github.com/nexussays/ble.net) but it doesn't support Linux or Mac.

**Additional context**
I'm currently using this library in a project to talk to an OBD-2 device that has BLE connectivity. It would be nice to have the same app modified for desktop so users can use their laptops. I've had numerous requests for Chromebook support, so the Linux desktop is very appealing.

remonh87:
@JoeGilkey  That is a nice feature request but not something we have on our roadmap on the short-term. However if you or someone else wants to provide support for desktop platforms we are more than happy to assist.

werediver:
In fact, our iOS implementation should be suitable (maybe with minor amendments) for macOS, but that's it. Neither Linux, nor Windows are supported or planned, as Remon mentioned.

Perhaps, the new [federated plug-in architecture](https://flutter.dev/docs/development/packages-and-plugins/developing-packages#federated-plugins) can be of use here, but it's not on our roadmap either.

ekuleshov:
So far this project demonstrated great activity in regards to maintaining code, providing fixes and applying community-provided pull requests. However lack of plans regarding support for desktop platforms is very discouraging. Especially macos and then windows platforms.

remonh87:
@ekuleshov I understand your request,  me and @werediver are actively maintaining iOS and Android which is already quite a big task that we do in our freetime. At this moment we do not have any bandwith to incorporate support for more platforms, however, if the community is willing to take on this task we are more then happy to support in reviewing prs and helping ppl get started

lukepighetti:
May I recommend adopting the new federated plugin architecture so that this repo is ready to start accepting work in these areas?

remonh87:
Yess it is on the road map #128 . Null-safety migration is almost done so I think moving to a federated structure would be a nice next step.

remonh87:
[...]

PR #401, opened 2021-10-05
Implement macOS support

juliansteenbakker:
#112 & #243

This is still a WIP, but for now scanning and discovering works!

Normally it would be the easiest to symlink the iOS code instead of duplicating it, however this is not possible because of the federated structure.

After some first tests it seems like connecting to a device is crashing somewhere. 

I'm making this PR public so that maybe other people can chime in and help improving macOS support.


juliansteenbakker:
This also needs some changes to reactive_ble_mobile package, so to get this to work locally you need to declare all package imports with path in pubspec.yaml like so:
```
    path: ../reactive_ble_platform_interface
```

remonh87:
@juliansteenbakker  first of all thank you for your contribution. I will look at it more thoroughly later this week.  I am not super happy about the code duplication, I think we should avoid this. What if we extract g the swift code out into its own package called `reactive_ble_swift`? Do you think symlink will work then?  Else I was thinking we can publish the swift implementation in a cocoa pod (however this always feels a bit of black magic to me).  What are you thoughts?

@werediver  can you also shine your light at this approach?

juliansteenbakker:
> I am not super happy about the code duplication, I think we should avoid this. 
Me neither.. I'm looking into some other options right now and update if i know more.

Taym95:
I will close this one since it is not active for a year now!

Issue #402, opened 2021-10-05
Backport to iOS 9


juliansteenbakker:
**Is your feature request related to a problem? (please describe)**
Right now the minimum iOS version is 11. For most people this isn't a problem however my implementation requires iOS 9 support. This is however achievable since the incompatibility right now is the usage of `CBManager` instead of `CBCentralManager`, and the functions of these two are the same. 

**Describe the solution you'd like**
You can use the following code to map `CBManager` to `CBCentralManager` :
```c
//#if __IPHONE_OS_VERSION_MAX_ALLOWED <= __IPHONE_9_3
#define CBManagerState CBCentralManagerState
#define CBManagerStateUnknown CBCentralManagerStateUnknown
#define CBManagerStateResetting CBCentralManagerStateResetting
#define CBManagerStateUnsupported CBCentralManagerStateUnsupported
#define CBManagerStateUnauthorized CBCentralManagerStateUnauthorized
#define CBManagerStatePoweredOff CBCentralManagerStatePoweredOff
#define CBManagerStatePoweredOn CBCentralManagerStatePoweredOn
//#endif
```

The only other function is a timer that needs to be converted.

```swift
    private func scheduleTaskTimeout(_ uniqueID: TaskQueue.Record.UniqueID, _ timeout: SubjectTask.Timeout) {
        if #available(iOS 10.0, *) {
            let timer = Timer.scheduledTimer(
                withTimeInterval: timeout.duration,
                repeats: false,
                block: papply(weak: self) { registry, _ in
                    registry.log("\(Controller.TaskSpec.tag) op timed out after \(timeout.duration) s, \(registry.tasks.count - 1) in queue")
                    timeout.handler()
                    registry.remove(uniqueID)
                }
            )
            scheduledTimeouts[uniqueID] = timer
        } else {
            //TODO: Implement 'old' timer function
            let timer = Timer.scheduledTimer(timeInterval: timeout.duration,
                                       target: self,
                                             selector: #selector(self.run(_:)),
                                     userInfo: nil,
                                      repeats: false)
            scheduledTimeouts[uniqueID] = timer
        }
        
    }
```

Once we've converted this, we can use this plugin on iOS 9 and higher. I tested it myself without the timer and i already got it working. The only thing i'm struggling with is how to convert the block from the `Timer.scheduledTimer` to a selector. Maybe someone has an idea?

Taym95:
[...]

PR #403, opened 2021-10-05
iOS 9 backport

juliansteenbakker:
Following up to #402, this PR enabled this package to be run on iOS 9.

There is one item that i need some help with which is the scheduleTaskTimeout function.
The existing function uses a block function, while the iOS 9 function needs a selector.


packages/reactive_ble_mobile/ios/Classes/Plugin/PluginController.swift
```diff
@@ -287,20 +282,6 @@ final class PluginController {
                 $0.characteristicUuids = (service.characteristics ?? []).map { characteristic in
                     Uuid.with { $0.data = characteristic.uuid.data }
                 }
-                $0.characteristics = (service.characteristics ?? []).map { characteristic in
-                    DiscoveredCharacteristic.with{
```
Taym95, line 291: Didn't you just remove `Expose characteristic properties` from #392 ?

juliansteenbakker: I think there was a sync problem with my local fork as it has just recently been added, i'll look into it tomorrow.

PR #433, opened 2021-11-02
Ensure the example is built against the local version of reactive_ble_mobile

robert-ancell:
Currently it seems to be pulling the version from pub.dev.

### DIFF
--- a/example/pubspec.yaml
+++ b/example/pubspec.yaml
@@ -11,8 +11,7 @@ dependencies:
   flutter:
     sdk: flutter
   flutter_lints: ^1.0.4
-  flutter_reactive_ble:
-    path: ../packages/flutter_reactive_ble
+  flutter_reactive_ble: ^5.0.1
   functional_data: ^1.0.0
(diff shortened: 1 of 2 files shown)


example/pubspec.yaml
```diff
@@ -11,8 +11,7 @@ dependencies:
   flutter:
     sdk: flutter
   flutter_lints: ^1.0.4
-  flutter_reactive_ble:
-    path: ../packages/flutter_reactive_ble
+  flutter_reactive_ble: ^5.0.1
```
remonh87, line 14: This should also point to the local version

remonh87 (approved):

example/pubspec.yaml
robert-ancell: The dependency_overrides below should ensure it does. I wasn't sure if both dependencies should be in dependency_overrides like I've made it, or the path should remain here as before. It felt to me more consistent to have both as overrides.

PR #437, opened 2021-11-04
Fix equality function for DiscoveredDevice

robert-ancell:


remonh87 (approved):
Good catch many thanks!

PR #448, opened 2021-11-14
Restore support for Xcode 12

ky1vstar:
Hi! This PR makes `flutter_reactive_ble` buildable and workable both in Xcode 12 and Xcode 13, which I would appreciate since our team not planning to migrate CI/CD to Xcode 13 for at least a couple of months. Thanks.

### DIFF
--- a/packages/reactive_ble_mobile/ios/Classes/Prelude/CoreBluetooth+Extensions.swift
+++ b/packages/reactive_ble_mobile/ios/Classes/Prelude/CoreBluetooth+Extensions.swift
@@ -1,3 +1,8 @@
+#if compiler(<5.5)
+import class CoreBluetooth.CBCharacteristic
+import class CoreBluetooth.CBService
+import class CoreBluetooth.CBPeripheral
+
 // Extensions below are supposed to backport API breaking changes
 // in CoreBluetooth framework which Apple introduced with Xcode 13:
 //
@@ -12,11 +17,6 @@
 //
 // - Note: This code compiles only when using Xcode 12 and below.
 // - SeeAlso: https://forums.swift.org/t/is-unowned-unsafe-t-weak-t-a-breaking-change/49917
-
-#if compiler(<5.5)
-import class CoreBluetooth.CBCharacteristic
-import class CoreBluetooth.CBService
-import class CoreBluetooth.CBPeripheral
 
 extension CBCharacteristic {
     @nonobjc
--- a/packages/reactive_ble_mobile/ios/Classes/Prelude/CoreBluetooth+Extensions.swift
+++ b/packages/reactive_ble_mobile/ios/Classes/Prelude/CoreBluetooth+Extensions.swift
@@ -1,0 +1,19 @@
+import class CoreBluetooth.CBCharacteristic
+import class CoreBluetooth.CBService
+import class CoreBluetooth.CBPeripheral
+
+#if compiler(<5.5)
+extension CBCharacteristic {
+    @nonobjc
+    var service: CBService? {
+        return value(forKey: #function) as? CBService
+    }
+}
+
+extension CBService {
+    @nonobjc
+    var peripheral: CBPeripheral? {
+        return value(forKey: #function) as? CBPeripheral
+    }
+}
+#endif
--- a/packages/reactive_ble_mobile/ios/Classes/Prelude/CoreBluetooth+Extensions.swift
+++ b/packages/reactive_ble_mobile/ios/Classes/Prelude/CoreBluetooth+Extensions.swift
@@ -2,17 +2,30 @@ import class CoreBluetooth.CBCharacteristic
 import class CoreBluetooth.CBService
 import class CoreBluetooth.CBPeripheral
 
+/// Extensions below are supposed to backport API breaking changes in CoreBluetooth
+/// framework which Apple introduced with Xcode 13:
+///
+/// ```
+/// + weak var peripheral: CBPeripheral? { get }
+/// - unowned(unsafe) var peripheral: CBPeripheral { get }
+/// ```
+///
+/// Thus this code shadows original property declarations in CoreBluetooth and changes their semantics
+/// from `unsafe non-optional` to `weak optional` to mimic Xcode 13 behaviour.
+///
+/// - Note: This code compiles only when using Xcode 12 and below.
+/// - SeeAlso: https://forums.swift.org/t/is-unowned-unsafe-t-weak-t-a-breaking-change/49917
 #if compiler(<5.5)
 extension CBCharacteristic {
     @nonobjc
-    var service: CBService? {
+    weak var service: CBService? {
         return value(forKey: #function) as? CBService
     }
 }
 
 extension CBService {
     @nonobjc
-    var peripheral: CBPeripheral? {
+    weak var peripheral: CBPeripheral? {
         return value(forKey: #function) as? CBPeripheral
     }
 }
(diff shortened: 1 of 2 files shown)


remonh87:
@werediver  can you check this one? It looks good to me but you are more proficient in Swift.

packages/reactive_ble_mobile/ios/Classes/Prelude/CoreBluetooth+Extensions.swift
```diff
@@ -0,0 +1,19 @@
+import class CoreBluetooth.CBCharacteristic
+import class CoreBluetooth.CBService
+import class CoreBluetooth.CBPeripheral
+
+#if compiler(<5.5)
+extension CBCharacteristic {
+    @nonobjc
+    var service: CBService? {
```
werediver, line 8: Both these properties are available since iOS 5.0. What you're trying to achieve with this?

Even if you have a reason to make this "shim", based on what (documentation) do you conclude that these objects actually have those named properties? This is to say it looks hacky / dirty.

werediver (changes requested):
I'd need to hear some solid explanation to approve this.

packages/reactive_ble_mobile/ios/Classes/Prelude/CoreBluetooth+Extensions.swift
ky1vstar: Hi! First of all, we need to figure out what exactly did Apple broke in Core Bluetooth APIs with Xcode 13. So they changed `- [CBCharacteristic service]` and `- [CBService peripheral]` semantic attribute from `unsafe` to `weak`. For Swift it requires now to add optional checks.
So what am I trying to achieve: 
- line 5: `compiler(<5.5)` means that included code should be compiled only in Xcode 13 and above (because only Xcode 13 do have Swift compiler version 5.5). Swift compiler version depends on Xcode version even so the project is compiled with Swift 4.2;
- line 9: #function == "service", I don't mind changing it to string literal;
- line 16: #function == "peripheral", I don't mind changing it to string literal;
- headers for `CBCharacteristic` class [here](https://developer.limneos.net/?ios=11.0&framework=CoreBluetooth.framework&header=CBCharacteristic.h) and `CBService` class [here](https://developer.limneos.net/?ios=11.0&framework=CoreBluetooth.framework&header=CBService.h) guarantee that above-mentioned classes will have required named properties in runtime;
- `@nonobjc` and internal declaration guarantee that we will not affect API behaviour anywhere outside our framework;
- converting properties from `unsafe` to `weak` is _okay_ since Apple intended them to be unsafe before Xcode 13;
- this fix doesn't required any changes in existing codebase.
Thus with this "shim" we have APIs which works identically both in Xcode 12 and Xcode 13, which I assume is pretty sweet.

ky1vstar:
@werediver I've replied in commit conversation

packages/reactive_ble_mobile/ios/Classes/Prelude/CoreBluetooth+Extensions.swift
werediver: @ky1vstar That's a pretty good explanation, I appreciate your effort. Some feedback:

> line 9: #function == "service", I don't mind changing it to string literal;
> line 16: #function == "peripheral", I don't mind changing it to string literal;

I actually like the current approach. Let's keep that as is.

> headers for `CBCharacteristic` class [here](https://developer.limneos.net/?ios=11.0&framework=CoreBluetooth.framework&header=CBCharacteristic.h) and `CBService` class [here](https://developer.limneos.net/?ios=11.0&framework=CoreBluetooth.framework&header=CBService.h) guarantee that above-mentioned classes will have required named properties in runtime;

You need to refer to the official documentation to derive any guarantees. That dump just proves certain properties of the _implementation_ in a particular OS build. Only the documented _public interface_ is reliable, not undocumented implementation details.

In this case, though, the subject properties _are_ a part of the public interface, so the official documentation says something very similar:

- [CBCharacteristic.service](https://developer.apple.com/documentation/corebluetooth/cbcharacteristic/1518728-service)
- [CBService.peripheral](https://developer.apple.com/documentation/corebluetooth/cbservice/1434334-peripheral)

> converting properties from `unsafe` to `weak` is okay since Apple intended them to be unsafe before Xcode 13;

You're not converting them to `weak`, though. The "shim" properties are now declared with a strong reference type. I'd rather prefer them to mimic the newer interface and use `weak` qualifier.

---

You see, the `unsafe` vs. `weak` is a subtle detail. This is worth putting in a comment or a doc-comment around the shim properties, otherwise it will be nigh impossible to figure out why that is needed some time later.

Please, reflect in the comments or doc-comments that:

- this code "backports" the API changes made in Xcode 13 so that they are available in Xcode 12
- namely, this code changes the type of the properties from unsafe non-optional to weak (optional)

---

So,

- [ ] with the addition of some comments or doc-comments
- [ ] and using `weak` qualifier in the shim properties declaration

this patch seems acceptable to me after your explanation.

ky1vstar: @werediver I'll provide comment a little bit later, but I do already user `weak` (optional) qualifier, don't I?

ky1vstar: @werediver thank you, I'll make appropriate changes.

ky1vstar: @werediver done

ky1vstar:
Should we also remove breaking changes notice from Readme?

werediver:
> Should we also remove breaking changes notice from Readme?

Good point. That would be nice, because thanks to your changes the library is again compatible with Xcode 12.

packages/reactive_ble_mobile/ios/Classes/Prelude/CoreBluetooth+Extensions.swift
```diff
@@ -0,0 +1,32 @@
...
+/// ```
+/// + weak var peripheral: CBPeripheral? { get }
+/// - unowned(unsafe) var peripheral: CBPeripheral { get }
+/// ```
+///
+/// Thus this code shadows original property declarations in CoreBluetooth and changes their semantics
+/// from `unsafe non-optional` to `weak optional` to mimic Xcode 13 behaviour.
+///
+/// - Note: This code compiles only when using Xcode 12 and below.
+/// - SeeAlso: https://forums.swift.org/t/is-unowned-unsafe-t-weak-t-a-breaking-change/49917
```
werediver, line 17: Thanks for adding this explanation, it looks good.

One nit-pick, though: a doc-comment (`///`) is associated with the declaration following it, but your explanation is not (instead it "belong" to the file scope: "Extensions below ..."), therefore I'd suggest to

- [ ] use a normal comment (`//`) in this case and separate it with an empty line from any following construct.

I hope my original wording ("comment or a doc-comment") was not too confusing. What I meant is that either would satisfy me, but you should choose the type that suits your particular explanation best. In this case, I believe, a normal comment should work best.

ky1vstar: I've decided to move comment to the top since it's related to the whole file

werediver: That is fine. But a doc-comment should be attached to a declaration, usually. Then it will be rendered in the quick-help panel and can be included in the generated HTML (or whatever) documentation, if anybody cares to generate it.

I'm not sure this arrangement associates the explanation with any reasonable entity (a declaration or the file), therefore I'm suggesting to make it a normal comment (unless you checked it and can confirm that the doc-comment is being associated with something reasonable).

werediver: @ky1vstar Perfect, thanks you.

werediver (approved):
A curious improvement. Thanks for you contribution, @ky1vstar

ky1vstar:
@werediver indeed. Thank you as well. When is it expected to be merged?

werediver:
Merged.

PR #453, opened 2021-11-17
Add support for writing to descriptors

ened:
Extends the plugins to support writing to specific descriptors. 
I have copied the approach from `QualifiedCharacteristic` to become a `QualifiedDescriptor`.

There are quite a few generated files (tests, protobuf) which are a bit tedious to handle. Appreciate if someone could point me to a developer docs page.

iOS implementation is still TBD.


remonh87:
thanks! so far it looks good

packages/reactive_ble_mobile/android/src/main/kotlin/com/signify/hue/flutterreactiveble/ble/ReactiveBleClient.kt
```diff
@@ -198,6 +198,33 @@ open class ReactiveBleClient(private val context: Context) : BleClient {
...
+                    is EstablishConnectionFailure -> {
+                        Single.just(
+                            CharOperationFailed(
+                                deviceId,
+                                "failed to connect ${connectionResult.errorMessage}"
+                            )
+                        )
+                    }
+                }
+            }.first(CharOperationFailed(deviceId, "Writechar timed-out"))
```
werediver, line 224: @remonh87 `CharOperation*` types are used multiple types in this method (and possibly in other places), while it's not exactly about characteristics, but rather about descriptors (which, admittedly, are associated with characteristics). Do you think it's fine or do you think a new set of types should be introduced?

I don't have a strong opinion on this, mainly because I'm not that familiar with the Android part. Just drawing your attention.

packages/reactive_ble_mobile/ios/Classes/ReactiveBle/Central.swift
```diff
@@ -275,6 +277,22 @@ final class Central {
...
+        value: Data,
+        descriptor qualifiedDescriptor: QualifiedDescriptor
+    ) throws {
+        let descriptor = try resolve(descriptor: qualifiedDescriptor)
+        
+        guard let characteristic = descriptor.characteristic else {
+            throw Failure.descriptorNotFound(qualifiedDescriptor)
+        }
+
+        guard let response = characteristic.service?.peripheral?.writeValue(value, for: descriptor)
```
werediver, line 291: @ened Is writing to a descriptor inherently a "without response" (without acknowledgement) operation? If so, then using `-WithoutResponse` is reasonable, but please clarify.

packages/reactive_ble_mobile/protos/bledata.proto
```diff
@@ -70,6 +70,16 @@ message WriteCharacteristicInfo {
     GenericFailure failure = 3;
 }
 
+message WriteDescriptorRequest {
+    DescriptorAddress descriptor = 1;
+    bytes value = 2;
```
werediver, line 75: @ened Do descriptors _always_ have 16-bit IDs and cannot have longer or shorter IDs?

packages/reactive_ble_mobile/pubspec.lock
```diff
@@ -350,9 +343,9 @@ packages:
   reactive_ble_platform_interface:
     dependency: "direct main"
     description:
-      name: reactive_ble_platform_interface
-      url: "https://pub.dartlang.org"
-    source: hosted
+      path: "../reactive_ble_platform_interface"
+      relative: true
+    source: path
```
werediver, line 348: Not sure this should be committed.

CC @remonh87

werediver (commented):
Looks pretty good. Good job, @ened 👍 

There is at least one "TODO" mark out there, so I consider this a work-in-progress for now.

ened:
@werediver @remonh87 944e2a9cae75e937431dc843693033462694d53f is not strictly part of this MR, but was needed for compatibility when including this package on iOS. Should it be separate?

packages/reactive_ble_mobile/ios/Classes/ReactiveBlePlugin.m
```diff
@@ -1,5 +1,10 @@
 #import "ReactiveBlePlugin.h"
+
+#if __has_include(<reactive_ble_mobile/reactive_ble_mobile-Swift.h>)
 #import <reactive_ble_mobile/reactive_ble_mobile-Swift.h>
+#else
+#import "reactive_ble_mobile-Swift.h"
+#endif
```
werediver, line 7: In which cases the original include doesn't work?

packages/reactive_ble_mobile/ios/reactive_ble_mobile.podspec
```diff
@@ -16,7 +16,10 @@ Bluetooth Low Energy (BLE) Flutter plug-in
   s.dependency 'SwiftProtobuf', '~> 1.0'
   s.ios.deployment_target = '11.0'
   s.osx.deployment_target = '10.13'
-  s.pod_target_xcconfig = { 'GCC_PREPROCESSOR_DEFINITIONS' => '$(inherited) GPB_USE_PROTOBUF_FRAMEWORK_IMPORTS=1' }
-  s.swift_version       = '4.2'
+  s.pod_target_xcconfig = {
+    'GCC_PREPROCESSOR_DEFINITIONS' => '$(inherited) GPB_USE_PROTOBUF_FRAMEWORK_IMPORTS=1',
+    'DEFINES_MODULE' => 'YES',
+    'VALID_ARCHS[sdk=iphonesimulator*]' => 'x86_64'
```
werediver, line 22: I wonder whether we'll need a new value here for the new M1 machines. But that can be resolved when we have input from people using them.

example/ios/Runner.xcodeproj/project.pbxproj
```diff
@@ -467,6 +468,7 @@
 				PRODUCT_BUNDLE_IDENTIFIER = com.signify.huereactivebleExample;
 				PRODUCT_NAME = "$(TARGET_NAME)";
 				SWIFT_OBJC_BRIDGING_HEADER = "Runner/Runner-Bridging-Header.h";
+				SWIFT_VERSION = 5.0;
```
werediver, line 471: Are change to this file important? This is the iOS "runner" project for the example app.

Also, this Swift version is different from the one declared in the pod-spec file for the library. Is 5.0 the highest Swift version the example app compiles with?

werediver (commented):
Some comments regarding 944e2a9cae75e937431dc843693033462694d53f.

werediver:
> Some comments regarding [944e2a9](https://github.com/PhilipsHue/flutter_reactive_ble/commit/944e2a9cae75e937431dc843693033462694d53f).

I suppose it's fine to include those changes in this pull request, but please go through all the comments already left by me.

Aside from those comments, the commit history of this pull request is dirty: contains merge commits (better rebase your feature branches), doesn't seem well structured and includes commit messages mentioning work-in-progress (WIP). If you are not going to clean the history up at some point (through interactive rebase), the pull request should be squashed before merging (can be done at the moment of merging through GitHub interface). It's okay, just don't be surprised.

ened:
> > Some comments regarding [944e2a9](https://github.com/PhilipsHue/flutter_reactive_ble/commit/944e2a9cae75e937431dc843693033462694d53f).
> 
> I suppose it's fine to include those changes in this pull request, but please go through all the comments already left by me.

OK & yes will do.

> Aside from those comments, the commit history of this pull request is dirty: contains merge commits (better rebase your feature branches), doesn't seem well structured and includes commit messages mentioning work-in-progress (WIP). If you are not going to clean the history up at some point (through interactive rebase), the pull request should be squashed before merging (can be done at the moment of merging through GitHub interface). It's okay, just don't be surprised.

Oh yes, I was fully expecting to squash all the commits before removing the `Draft` label.

Thank you for the reviews.

remonh87:
I can test the project setup on the m1 machine (I have one) if needed.

example/ios/Runner.xcodeproj/project.pbxproj
ened: I think this was purely done to reduce warnings in Xcode and declare the actually used version. 

5.5 is probably becoming 5.0 (when declared in .podspec) as this is more about the language level than a specific version.

packages/reactive_ble_mobile/ios/Classes/ReactiveBlePlugin.m
ened: Reason this is changed is to improve compatibility with host projects which do not use Swift or `use_frameworks!` (or use static frameworks - I can't remember).

There was a lengthy issue on the Flutter GH repo and all projects with native were migrated to this pattern.

packages/reactive_ble_mobile/ios/Classes/Plugin/PluginController.swift
```diff
@@ -491,6 +497,66 @@ final class PluginController {
...
+              }) else {
+            return
+        }
+        
+        if let _ = characteristic.descriptors?.first(where: { $0.uuid == descriptor.id }) {
+            let result = performWriteDescriptorWithoutResponse(central: central, args: args, descriptor: descriptor)
+            completion(.success(result))
+        } else {
+            do {
+                try central.discoverDescriptors(
```
ened, line 528: @werediver @remonh87 would like your opinion if this is the correct approach. I think there is no need to discover all characteristics from the get go. It may also imply a API change to declare what to discover. 

The code currently discovers on-demand but may change the library behavior from what the use expects?

packages/reactive_ble_mobile/ios/reactive_ble_mobile.podspec
ened: I was testing this on a M1 Air Simulator OK - hence I would assume no. This is one of the build settings (the define module part as well) needed for greater compatibility.

werediver:
Reminding of my [earlier question](https://github.com/PhilipsHue/flutter_reactive_ble/pull/453#discussion_r754406740):

> @ened Do descriptors _always_ have 16-bit IDs and cannot have longer or shorter IDs?

`CBDescriptor.id` has type `CBUUID` and that thing can contain IDs of different length, typically 128, 32, or 16 bits, while strictly 16 bits is reserved in Protobuf structure.

I'd suggest to find a suitable example in existing Protobuf structures and reuse it.

- [ ] Rework or clarify Protobuf descriptor ID format

packages/reactive_ble_mobile/ios/Classes/Plugin/PluginController.swift
werediver: Characteristic discovery works differently on iOS and Android. On iOS a client can specify a list of services and characteristics to discover when connecting to a device (the client is only allowed to interact with the discovered characteristics later on), while on Android no discovery is triggered beforehand (I think it's happening on the fly then, but entirely sure; CC @remonh87).

In terms of design, a simple solution is to extend the existing discovery mechanism to also discover descriptors, as specified when connecting. That would be consistent and quite efficient.

Discovering descriptors one by one is definitely not as efficient (the more descriptors a client uses, the larger the overhead is). This is also not consistent with how the library works with characteristics (on iOS).

ened: @werediver go you regarding the differences - and will think how to extend the initial discovery. It does mean the existing iOS discovery code will get another layer (currently counts how many services are left to discover, this will be extended with the number of characteristics to explore).

packages/reactive_ble_mobile/ios/Classes/ReactiveBle/Central.swift
ened: Writing to descriptors works via https://developer.apple.com/documentation/corebluetooth/cbperipheral/1519107-writevalue . There is no response per se, but the docs at https://developer.apple.com/documentation/corebluetooth/cbperipheraldelegate/1519062-peripheral state that a delegate method will be called once the write is completed. If it fails, the error is given as well. Do you think the plugin should implement that delegate and return a value only once that is done?

`WithoutResponse` still applies based on that logic, but I am wondering what the plugin should do.

ened:
> Reminding of my [earlier question](https://github.com/PhilipsHue/flutter_reactive_ble/pull/453#discussion_r754406740):
> 
> > @ened Do descriptors _always_ have 16-bit IDs and cannot have longer or shorter IDs?
> 
> `CBDescriptor.id` has type `CBUUID` and that thing can contain IDs of different length, typically 128, 32, or 16 bits, while strictly 16 bits is reserved in Protobuf structure.
> 
> I'd suggest to find a suitable example in existing Protobuf structures and reuse it.
> 
> * [ ]  Rework or clarify Protobuf descriptor ID format

At the moment, the code defines `DescriptorAddress` like this:

```
message DescriptorAddress {
    string deviceId = 1;
    Uuid serviceUuid = 2;
    Uuid characteristicUuid = 3;
    Uuid descriptorUuid = 4;
}
```

and `Uuid` is defined as:

```
message Uuid {
    bytes data = 1;
}
```

So that looks good?

werediver:
> > Reminding of my [earlier question](https://github.com/PhilipsHue/flutter_reactive_ble/pull/453#discussion_r754406740):
> > > @ened Do descriptors _always_ have 16-bit IDs and cannot have longer or shorter IDs?
> > 
> > 
> > `CBDescriptor.id` has type `CBUUID` and that thing can contain IDs of different length, typically 128, 32, or 16 bits, while strictly 16 bits is reserved in Protobuf structure.
> > I'd suggest to find a suitable example in existing Protobuf structures and reuse it.
> > 
> > * [x]   Rework or clarify Protobuf descriptor ID format
> 
> At the moment, the code defines `DescriptorAddress` like this:
> 
> ```
> message DescriptorAddress {
>     string deviceId = 1;
>     Uuid serviceUuid = 2;
>     Uuid characteristicUuid = 3;
>     Uuid descriptorUuid = 4;
> }
> ```
> 
> and `Uuid` is defined as:
> 
> ```
> message Uuid {
>     bytes data = 1;
> }
> ```
> 
> So that looks good?

Oh, sure, it looks good. I misread a piece of protobuf specification, sorry.

packages/reactive_ble_mobile/ios/Classes/ReactiveBle/Central.swift
werediver: > Do you think the plugin should implement that delegate and return a value only once that is done?

If waiting for the events delivered to the delegate adds anything, it might be a potential improvement, but at of now, I'd say, this method should be implemented similar to https://github.com/PhilipsHue/flutter_reactive_ble/blob/d3d278163e67978c3df9ab2fe7e890f840cb23fd/packages/reactive_ble_mobile/ios/Classes/ReactiveBle/Central.swift#L264

As long as they match, it should be enough.

packages/reactive_ble_mobile/ios/Classes/Plugin/PluginController.swift
```diff
@@ -491,6 +497,66 @@ final class PluginController {
...
+        }
+        
+        guard let descriptor = QualifiedDescriptorIDFactory().make(from: args.descriptor)
+        else {
+            completion(.failure(PluginError.invalidMethodCall(method: name, details: "descriptor, characteristic, service, and peripheral IDs are required").asFlutterError))
+            return
+        }
+        
+        guard let peripheral = central.activePeripherals[descriptor.peripheralID],
+              let service = peripheral.services?.first(where: { service in
```
corecode, line 514: how will `.first` work with devices that provide several instances of the same service (UUID)?  Same for several instances of the same characteristic UUID.

ened: Devices can provide multiple services with the same UUID? I have to admit that's a part of the spec I do not know about!

Technically this May be a more fundamental problem in this package then, looking as QualifiedCharacteristic for example, a full lookup is done using the UUIDs but not via index?

albydarned:
I need to be able to read a characteristic's user descriptor... Does this PR just include writes?

packages/reactive_ble_mobile/ios/Classes/Plugin/PluginController.swift
spkersten: In case anyone wants to continue with this in the future. Have a look at https://github.com/PhilipsHue/flutter_reactive_ble/pull/776 where the problem of multiple characteristics of services with the same id has been addressed. Something similar would be needed here.

Issue #476, opened 2021-12-18
Example App failes to start


ride4sun:
**Describe the bug**

When I run the example app I get this error:

Running "flutter pub get" in example...

```
Warning: You are using these overridden dependencies:
! flutter_reactive_ble 5.0.2 from path ..\packages\flutter_reactive_ble
! reactive_ble_mobile 5.0.2 from path ..\packages\reactive_ble_mobile
Launching lib\main.dart on motorola edge 2021 in debug mode...
Running Gradle task 'assembleDebug'...
√  Built build\app\outputs\flutter-apk\app-debug.apk.
I/flutter (10254): Observatory server failed to start after 1 tries
I/flutter (10254): Observatory server failed to start after 2 tries
I/flutter (10254): Observatory server failed to start after 3 tries
I/flutter (10254): Observatory server failed to start after 4 tries
I/flutter (10254): Observatory server failed to start after 5 tries
I/flutter (10254): Observatory server failed to start after 6 tries
I/flutter (10254): Observatory server failed to start after 7 tries
I/flutter (10254): Observatory server failed to start after 8 tries
I/flutter (10254): Observatory server failed to start after 9 tries
I/flutter (10254): Observatory server failed to start after 10 tries
I/flutter (10254): Observatory server failed to start after 11 tries
I/flutter (10254): Could not start Observatory HTTP server:
I/flutter (10254): SocketException: Failed to create server socket (OS Error: Operation not permitted, errno = 1), address = 127.0.0.1, port = 0
I/flutter (10254): #0      _NativeSocket.bind (dart:io-patch/socket_patch.dart:988:7)
I/flutter (10254): <asynchronous suspension>
I/flutter (10254): #1      Server.startup.poll (dart:vmservice_io/vmservice_server.dart:425:19)
I/flutter (10254): <asynchronous suspension>
I/flutter (10254): #2      Server.startup (dart:vmservice_io/vmservice_server.dart:437:13)
I/flutter (10254): <asynchronous suspension>
I/flutter (10254): 
```

Platform and Flutter Version:
```
C:\Projects>flutter doctor
Doctor summary (to see all details, run flutter doctor -v):
[√] Flutter (Channel stable, 2.8.1, on Microsoft Windows [Version 10.0.19043.1415], locale en-US)
[√] Android toolchain - develop for Android devices (Android SDK version 30.0.3)
[√] Chrome - develop for the web
[√] Android Studio (version 2020.3)
[...]

PR #477, opened 2021-12-21
Add INTERNET permission to example app

Taym95:
#476


remonh87 (approved):
thanks. DOn't know why the CI is failing though

Taym95:
@remonh87 can I see what is the CI error output?

remonh87 (commented):
@Taym95  I assume we need this for debugging right? Shouldn't we add it only to the debug manifest in that case so it wouldn't confuse users? Ble does not need need internet

PR #506, opened 2022-01-26
Fix Android Permissions

rickcasson:
I've observed that Android 12 devices that use `ACCESS_COARSE_LOCATION` and `ACCESS_FINE_LOCATION` for purposes other than Bluetooth scanning will encounter issues when `maxSdkVersion` is enforced. 

`uses-permission-sdk-23` also makes scanning impossible for devices with SDK 21 and 22 (Android 5 and 5.1.1)

As pointed out in #477, the `INTERNET` permission is needed in the example app, but I also removed all the Bluetooth related permissions as they will get inherited from the plugin

### DIFF
--- a/example/android/app/src/main/AndroidManifest.xml
+++ b/example/android/app/src/main/AndroidManifest.xml
@@ -1,27 +1,7 @@
 <manifest xmlns:android="http://schemas.android.com/apk/res/android"
     package="com.signify.hue.reactivebleexample">
 
-    <!-- required for API 18 - 30 -->
-    <uses-permission
-            android:name="android.permission.BLUETOOTH"
-            android:maxSdkVersion="30" />
-    <uses-permission
-            android:name="android.permission.BLUETOOTH_ADMIN"
-            android:maxSdkVersion="30" />
-
-    <!-- required for API 23 - 30 -->
-    <uses-permission-sdk-23
-            android:name="android.permission.ACCESS_COARSE_LOCATION"
-            android:maxSdkVersion="30" />
-    <uses-permission-sdk-23
-            android:name="android.permission.ACCESS_FINE_LOCATION"
-            android:maxSdkVersion="30" />
-
-    <!-- API 31+ -->
-    <uses-permission android:name="android.permission.BLUETOOTH_CONNECT" />
-    <uses-permission
-            android:name="android.permission.BLUETOOTH_SCAN"
-            android:usesPermissionFlags="neverForLocation" />
+    <uses-permission android:name="android.permission.INTERNET"/>
 
     <application
         android:label="flutter_reactive_ble_example"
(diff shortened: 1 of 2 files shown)


Taym95:
@PieterAelse can you check this? He removed quite some stuff from your Android 11 support

example/android/app/src/main/AndroidManifest.xml
```diff
@@ -1,27 +1,7 @@
...
-            android:maxSdkVersion="30" />
-    <uses-permission-sdk-23
-            android:name="android.permission.ACCESS_FINE_LOCATION"
-            android:maxSdkVersion="30" />
-
-    <!-- API 31+ -->
-    <uses-permission android:name="android.permission.BLUETOOTH_CONNECT" />
-    <uses-permission
-            android:name="android.permission.BLUETOOTH_SCAN"
-            android:usesPermissionFlags="neverForLocation" />
```
Taym95, line 24: I don't think these permissions need to be removed! are needed for any app that uses android 12!

example/android/app/src/main/AndroidManifest.xml
```diff
@@ -1,27 +1,7 @@
...
+    <uses-permission android:name="android.permission.INTERNET"/>
```
Taym95, line 4: Is needed for debugging? I think you need to add it only to the debug manifest

Taym95 (changes requested):

example/android/app/src/main/AndroidManifest.xml
rickcasson: This is for the example app, not the plugin itself. When the project is built, it will generate a manifest using the permissions it inherits for all its dependencies (including this plugin). This change was just to remove redundancy.

example/android/app/src/main/AndroidManifest.xml
rickcasson: Yes, it is needed for running the example app in debug mode. But this is just the example app, so it won't impact the plugin's required permissions.

example/android/app/src/main/AndroidManifest.xml
Taym95: Yes but the example app shows how to use the lib in a real-life app, if you don't add these permissions (with flutter-permission-handler for example ) permissions will never be handled, if you want to contribute will be better to add an example of how to use flutter-permission-handler with the library for example rather than remove permissions from example app!

rickcasson: A real-life app and the example app will inherit these exact same permissions from the plugin so they actually don't actually need to be in the app's manifest (look in build/app/intermediates/merged_manifests after you build the example app). But I'm fine dropping these changes as fixing the plugin's manifest is more important to me anyway. Unfortunately I've realized for these changes to work, RxAndroidBle will need to similarly modify its AndroidManifest as its permissions get inherited too...

vincent-iQontrol:
I have also noticed the issue.  Because I need the location permission for Wifi info.

Taym95:
@rickcasson thanks a lot for you contributing, I will close this PR since #524 is addressing the same issue with other approach

PR #519, opened 2022-02-17
Deinitialize in onDetachedFromEngine on Android

Rexios80:
Without this, devices stay connected to detached native code and behave unexpectedly when reattached to flutter

### DIFF
--- a/packages/reactive_ble_mobile/android/src/main/kotlin/com/signify/hue/flutterreactiveble/PluginController.kt
+++ b/packages/reactive_ble_mobile/android/src/main/kotlin/com/signify/hue/flutterreactiveble/PluginController.kt
@@ -71,6 +71,11 @@ class PluginController {
         bleStatusChannel.setStreamHandler(bleStatusHandler)
     }
 
+    internal fun deinitialize() {
+        scandevicesHandler.stopDeviceScan()
+        deviceConnectionHandler.disconnectAll()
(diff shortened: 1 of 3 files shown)


packages/reactive_ble_mobile/android/src/main/kotlin/com/signify/hue/flutterreactiveble/PluginController.kt
```diff
@@ -71,6 +71,11 @@ class PluginController {
         bleStatusChannel.setStreamHandler(bleStatusHandler)
     }
 
+    internal fun deinitialize() {
```
Taym95, line 74: can be private

Taym95 (changes requested):

Rexios80:
This crashes if there aren't any devices connected

Rexios80:
Fixed that. Lateinit variables are dangerous.

Also I used the current instance of the plugin as the method handler. For this plugin it doesn't look like that changes anything, but it matches the new flutter plugin template and can prevent issues in the future.

remonh87 (approved):
yes much better thanks!!

Taym95 (approved):

Rexios80:
Any idea when this will be released?

Taym95:
@Rexios80 we will release a new version ASAP

PR #524, opened 2022-02-25
Removed unsafe manifest permissions and updated readme 

Rexios80:
These permissions should be handled by the users of this plugin. There are too many edge cases with manifest merge issues for this plugin to handle them itself.


README.md
```diff
@@ -211,32 +211,16 @@ RxJavaPlugins.setErrorHandler { throwable ->
...
-* BLUETOOTH : allows apps to connect to a paired Bluetooth device
-* BLUETOOTH_ADMIN: allows apps to discover and pair Bluetooth devices
-* BLUETOOTH_SCAN: add this permission with `tools:node="remove"` to remove it from the merged manifest.
-``` 
-<uses-permission
-            android:name="android.permission.BLUETOOTH_SCAN"
-            android:usesPermissionFlags="neverForLocation" tools:node="remove" />
+You need to add the following permissions to your AndroidManifest.xml file:
+```xml
+<uses-permission android:name="android.permission.BLUETOOTH_SCAN" android:usesPermissionFlags="neverForLocation" />
```
Taym95, line 216: Thanks for your contribution, BLUETOOTH_SCAN was added only in API 31, why did you remove distinguish between API 31,30?

Taym95 (changes requested):

README.md
Rexios80: That's from the official documentation: https://developer.android.com/guide/topics/connectivity/bluetooth/permissions

packages/reactive_ble_mobile/android/src/main/AndroidManifest.xml
```diff
@@ -9,17 +9,6 @@
         android:name="android.permission.BLUETOOTH_ADMIN"
         android:maxSdkVersion="30" />
 
-    <!-- required for API 23 - 30 -->
```
PieterAelse, line 12: So this is removed to prevent conflicts in apps using the plugin, but why should `android.permission.BLUETOOTH_CONNECT` be kept then? As `android.permission.BLUETOOTH_SCAN` is also removed below?

Rexios80: Because `BLUETOOTH_CONNECT` doesn't have any possible modifiers. `BLUETOOTH_SCAN` has the optional `neverForLocation` modifier. I guess maybe there's a small chance someone using this plugin would only want to scan and not connect maybe? But that would probably be a very niche use-case.

Rexios80:
@PieterAelse Can't you squash them when you merge? I'm pretty sure I'd have to make a new PR to squash them myself.

packages/reactive_ble_mobile/android/src/main/AndroidManifest.xml
PieterAelse: Usage will then become a bit weird.. Because users of the plugin **have to** specify the `SCAN` in their manifest, but not per-se the `CONNECT` because that's already done by the plugin. 

IMO I think we should then also not do `CONNECT` and update the readme that they'll have to add both `SCAN` and `CONNECT`/`ADVERTISE` (if they use it) to their manifest. But lemme check what the other maintainers think: 
@Taym95 @remonh87 Your opinion on this?

Taym95: I agree with @PieterAelse let's be consistent

Rexios80:
I noticed that the root README and the README for flutter_reactive_ble are very similar, but not the same. Should this be dealt with? Also the permissions and proguard parts of the readme should really be close to the top so users don't miss them. Is it okay if I move them up there?

Taym95:
The one in flutter_reactive_ble is outdated and can you update it and move permissions after `Usage` please?

Rexios80:
I was thinking maybe the root readme should be linked to the flutter_reactive_ble readme somehow to avoid having to copy/paste changes. Or the root readme could have only the contribution guidelines and leave everything else in the flutter_reactive_ble readme.

Rexios80:
I updated the readmes to what I would do for one of my packages. If you want me to change it just let me know.

Rexios80:
I was looking at the ProGuard configuration, and it looks like the readme information for it should be irrelevant since you have consumerProguardFiles set in the build.gradle file. Should the proguard information be removed from the readme to avoid confusion?

PieterAelse (approved):

PieterAelse:
I'm fine with the manifest changes. I've added @remonh87  as a reviewer (next to @Taym95 ) to decide on the changes in the Readme files. I think they'll also influence what is shown on pub.dev right?

For merging; let's squash indeed.

Rexios80:
I can put it back but I don't like having duplicated stuff that can easily get out of sync. There is precedent for doing similar things with readmes:

https://github.com/flutter/plugins
https://github.com/fluttercommunity/plus_plugins
https://github.com/cph-cachet/flutter-plugins

We could add a table of contents like these repos do

Taym95 (approved):

PR #542, opened 2022-03-28
update README.md to include Android SDK 31+ Location Permission FAQ example

adamwendel:
I found the documentation on how to do this very lacking and confusing to follow as a flutter-first developer. As it can be typical to use bluetooth and location in-conjunction with each other I propose adding a section to the FAQ part of README.md to help those that might use the library in the future and have the same question.

### DIFF
--- a/README.md
+++ b/README.md
@@ -249,6 +249,17 @@ Before you are able to execute BLE operations the BLE-stack of the device makes 
 
 This will prevent issues like [#147](https://github.com/PhilipsHue/flutter_reactive_ble/issues/147).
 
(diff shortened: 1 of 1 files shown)


PieterAelse:
With which version of the codebase did you experience this? As we very recently merged changes related to permissions and the Readme, see here: https://github.com/PhilipsHue/flutter_reactive_ble/pull/524

adamwendel:
Ah thank you for that. I didn't see those recent changes. I was using the latest from pub.dev and finally got around to making a PR and didn't see the other PR that will have those breaking changes to have end-developers put in permissions. Thanks for your time and this PR can be closed without merge

PR #544, opened 2022-03-29
Implement web/mac/windows/linux support 

rohitsangwan01:
Web Support with almost all Functionalities working , need some improvements though 

Unimplemented Methods : 
clearGattCache
requestMtuSize
requestConnectionPriority

Here is a demo of example app running without any changes ( just added web folder) 

https://user-images.githubusercontent.com/59526499/160696123-322f2200-daa9-4276-85e9-76d3969a41df.mp4


rohitsangwan01:
@remonh87  project is building fine for me in Android +Ios+Web , not sure why these checks are failing
Also is there any way , that i can create my own package, something like
flutter_reactive_ble_web_extension
And use it in project something like this

FlutterReactiveBle ble =  FlutterReactiveBle(
WebExtension: FlutterReactiveBleWeb,
WindowsExtension : .......
)

remonh87:
@rohitsangwan01  we would prefer if you use a separate package and publish it yourself :) . We of course can add your package to the readme. @Taym95  what do you think?

rohitsangwan01:
@remonh87 yes but my current project is totally based on flutter_reactive_ble , and now i have to add other platforms support as well , so i wanted a way somehow , add support for different platforms without touching code , as Android/Ios is already stable ,and this web and Desktop (Windows + mac+linux) extensions i added to this package are already available in other packages
am contributing to [quick_blue](https://pub.dev/packages/quick_blue) package , so here i just added quick_blue to do the native tasks , and for web am using [flutter_web_bluetooth](https://pub.dev/packages/flutter_web_bluetooth)
, so i guess no sense in uploading new packages right, but if i can somehow extend flutter_reactive_ble , by using my github repo or something , 
i know last option is to use my fork , but if possible , i wanted to use
this kind of option i, in which , for Android/Ios i can use official flutterReactiveBLe , and for other platforms i can use ReactiveBlePlatformFactory of my fork  ,and even  it will be helpfull for other peoples too  if they want to extend their platform support , or wanted to test
or another last option in my case is to create two projects , one with official package for Mobile , and one with Desktop/Web support only using my fork
any suggestions ,if there is something else i can try , without creating two projects

Taym95:
> @rohitsangwan01 we would prefer if you use a separate package and publish it yourself :) . We of course can add your package to the readme. @Taym95 what do you think?

Totally agree with @remonh87, we want to keep this repo for IOS/Android, and we are open to endorsing plugins for other platforms.

rohitsangwan01:
@remonh87 @Taym95 okk great, we can go with endorsing

Nolence:
Just a random passerby but I think it makes sense for flutter packages to not separate out platforms. Flutter itself is trying to be very cross platform by targeting most anything (whether this is good or bad) so I don't think it would be unusual to include web/mac etc. A table of platform features and yet-to-be-implemented's in the Readme is usually how I've seen other packages go about similar situations.

nobodyguy:
> > @rohitsangwan01 we would prefer if you use a separate package and publish it yourself :) . We of course can add your package to the readme. @Taym95 what do you think?
> 
> Totally agree with @remonh87, we want to keep this repo for IOS/Android, and we are open to endorsing plugins for other platforms.

I mean cross-platform is the whole point of Flutter.

Issue #384, opened 2021-09-27
Re-Enable notifications on reconnection to device


rowdy15:
**Describe the bug**
Basically the notifications are not re-enabled on reconnection to the ble device.

This is supported by the below code where on first setup of the .connectToDevice() listener. I recieve the below output on first connection when subscribing to the three characteristics.
```
D/BluetoothGatt( 3822): setCharacteristicNotification() - uuid: f3641401-00b0-4240-ba50-05ca45bf8abc enable: true
D/BluetoothGatt( 3822): setCharacteristicNotification() - uuid: f3641403-00b0-4240-ba50-05ca45bf8abc enable: true
D/BluetoothGatt( 3822): setCharacteristicNotification() - uuid: 00002a19-0000-1000-8000-00805f9b34fb enable: true
```
on disconnection I get the following.
```
D/BluetoothGatt( 5669): setCharacteristicNotification() - uuid: f3641401-00b0-4240-ba50-05ca45bf8abc enable: false
D/BluetoothGatt( 5669): setCharacteristicNotification() - uuid: f3641403-00b0-4240-ba50-05ca45bf8abc enable: false
D/BluetoothGatt( 5669): setCharacteristicNotification() - uuid: 00002a19-0000-1000-8000-00805f9b34fb enable: false
```

But unfortunately on reconnection when I go back into range of the BLE device (and it reconnects) I don't get a reenabling of the the notifications of those same characteristics.

Below is the complete output of during these events.

```
D/BluetoothGatt(17897): setCharacteristicNotification() - uuid: f3641401-00b0-4240-ba50-05ca45bf8abc enable: true
D/BluetoothGatt(17897): setCharacteristicNotification() - uuid: f3641403-00b0-4240-ba50-05ca45bf8abc enable: true
D/BluetoothGatt(17897): setCharacteristicNotification() - uuid: 00002a19-0000-1000-8000-00805f9b34fb enable: true
I/flutter (17897): periodic clock updates enabled in 54 seconds
D/BluetoothGatt(17897): onClientConnectionState() - status=0 clientIf=6 device=EF:2C:0C:2E:78:1A
D/BluetoothGatt(17897): setCharacteristicNotification() - uuid: f3641401-00b0-4240-ba50-05ca45bf8abc enable: false
D/BluetoothManager(17897): getConnectionState()
D/BluetoothManager(17897): getConnectedDevices
D/BluetoothGatt(17897): close()
[...]

PR #588, opened 2022-07-07
Cancel subscription when a disconnect event has been thrown

sjoerd0301:
How to replicate:

[Android only]

- connect to device
- enable notifications
- power device off without disconnecting. (or walk out of range)
- wait for disconnect event from plugin
- power on device (or walk in range when you walked out of range in step 3)
- scan and connect to device
- enable notifications -> This should give an error the first time after disconnect. (EstablishedConnectionUpdate failure device disconnected MAC="...." etc)
- cancel subscription
- enable notifications again -> works

Somehow the EstablishedconnectionUpdate gets lost/cached/whatever when a device disconnects.
This PR cancels the characteristic notificationstream on the android native side.

Possible fix for #384


sjoerd0301:
@remonh87 @Taym95 @werediver?

example/pubspec.lock
```diff
@@ -205,8 +205,8 @@ packages:
   flutter_reactive_ble:
     dependency: "direct main"
     description:
-      path: "../packages/flutter_reactive_ble"
-      relative: true
+      path: "/home/sjoerd/source/flutter_reactive_ble/example/../packages/flutter_reactive_ble"
```
Taym95, line 208: I don't think you username should be here

Taym95 (changes requested):

example/pubspec.lock
werediver: There are more entries like this below.

sjoerd0301: .lock files should be ignored in the gitignore.
.lock files will be generated on a new build.

Please take a look at the 'packages/reactive_ble_mobile/android/src/main/kotlin/com/signify/hue/flutterreactiveble/channelhandlers/CharNotificationHandler.kt' file.

sjoerd0301: Maybe It's a good idea to update the .gitignore in the project root?

Taym95: pubspec.lock need to be committed

werediver: Please read [What not to commit](https://dart.dev/guides/libraries/private-files).

Lock-files for libs, as usual, should not be committed, but for the apps they should be. There should be no lock-files for the libs in the repo, but for the example app there should be.

@sjoerd0301 Make sure you comply with the guidelines regarding this.

sjoerd0301: Deleted .lock files.
Please take a look at the kotlin code, as that _does_ fix a problem

werediver: You deleted more lock-files, then necessary. The example app is an _app_, lock-files for the apps are to be committed.

A pull request gets merged in its entirety, if one file is good, but others are not, it still doesn't work.

packages/reactive_ble_mobile/android/src/main/kotlin/com/signify/hue/flutterreactiveble/channelhandlers/CharNotificationHandler.kt
```diff
@@ -36,7 +37,14 @@ class CharNotificationHandler(private val bleClient: com.signify.hue.flutterreac
             .subscribe({ value ->
                 handleNotificationValue(request.characteristic, value)
             }, {
-                handleNotificationError(request.characteristic, it)
+                when (it) {
+                    is BleDisconnectedException -> {
+                        (subscriptionMap[request.characteristic])?.dispose()
```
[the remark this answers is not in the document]
werediver: @remonh87 If you're up to providing any input on this, it would be very welcome 🙂

werediver:
@sjoerd0301 Thank you for the update. Do you believe you tested your changes sufficiently? If yes, this looks good to me.

CC @Taym95 @remonh87

rfuerst87:
Hi all

Any chances this gets merged? We currently have issues when reconnecting to a device that was out of range. Subscribing to characteristics sometimes throws ` CharacteristicValueUpdateError.unknown` errors. Issue #609 describes our case quite well. Unfortunately there is issues with the proposed solution as well. I found that delays are considered only on the initial connect but not on any subsequent reconnects unless it is set to something above 500 ms. I don't really like this workaround... It is somehow arbitrary and it might work on one mobile but not on another. This problem is especially nasty as we see it in release builds only.

Anyway, I just stumbled upon this MR which seems to solve all our issues. We can even get rid of delays. Obviously I'm very interested in a merge. Is there anything I can do here to speed up the process?

Many thanks for your efforts!

meuhle:
@sjoerd0301 I've tried to copy your changes to the library but seems to work same as the old way, maybe I'm doing something wrong with the dependencies, but I've imported flutter_reactive_ble inside my project and renamed and added to pubspec.yaml of the project the new path, but to mantain the chnages I also copied the reactive_ble_mobile and adjusted its pubspec.yaml and then inside of its foder i copied the reactive_ble_platform_interface , cuold you help me to reproduce your case?

rfuerst87:
@meuhle you can directly point your dependencies in pubspec.yaml to @sjoerd0301's fork:

```yaml
flutter_reactive_ble:
  git:
    url: https://github.com/sjoerd0301/flutter_reactive_ble.git
    ref: bug/cancel_char_notif_on_disconnect
    path: packages/flutter_reactive_ble
```

additionally you will need to override dependencies, as otherwise wrong dependencies get pulled in:

```yaml
dependency_overrides:
  reactive_ble_mobile:
    git:
      url: https://github.com/sjoerd0301/flutter_reactive_ble.git
      ref: bug/cancel_char_notif_on_disconnect
      path: packages/reactive_ble_mobile
    git:
      url: https://github.com/sjoerd0301/flutter_reactive_ble.git
      ref: bug/cancel_char_notif_on_disconnect
      path: packages/reactive_ble_platform_interface
```

By the way: Is there any updates on this MR? We're still relying on a custom fork on this repo just because of this one issue. Is there anything that that speaks against merging? @Taym95 @werediver @remonh87

remonh87:
For some reason I didn't receive any notifications for this repo and I was not working with ble last year. I looked into the PR and it seems good to me, good catch!

remonh87 (approved):

farr64:
Welcome back, @remonh87

Taym95 (approved):

Taym95:
@sjoerd0301 lets rebase and merge it!

Taym95:
I will close this one and clone it in #769

Issue #571, opened 2022-06-07
Can't find peripherals that implement extended advertising


BikeBr0:
**Describe the bug**
Since our peripheral switched to using extended advertising, I have not been able to find the peripheral when performing a scan. It was previously working when the peripheral used standard advertising. 

**To Reproduce**
1. Scan for peripherals that implement extended advertising  `_ble.scanForDevices(withServices: withServices)`

**Expected behavior**
To find the peripheral

**Smartphone / tablet**
 - Google Pixel 6 (Android 12) 
 - Google Nexus 6P (Android 8.1.0)
 - flutter_reactive_ble: ^5.0.2
 - Flutter version 3.0.1
 - Dart version 2.17.1

**Peripheral device**
- nRF52840-DK
- Does it run a custom firmware / software: yes

**Additional context**
This is an android specific bug as both advertising models work on iOS. Also, peripherals using extended advertising do appear in the scan results of the nRF Connect App on the Android Pixel 6.

BikeBr0:
Further to the information provided above, we have tried using sample firmware provided with the nRF52840-DK and configured it to use Bluetooth 5 Advertising Extension. The peripheral shows up on nRF Connect on the Google Pixel 6 but not in my custom Flutter App. Is this not supported by flutter_reactive_ble?

rift-labs-developer:
Im having the same problem. It seems RxAndroidBle does not support extended advertising yet (https://github.com/dariuszseweryn/RxAndroidBle/issues/774 )

danielstuart14:
https://github.com/PhilipsHue/flutter_reactive_ble/pull/594

danielstuart14:
@Taym95 is there any ETA for a release with the fix?

Taym95:
@danielstuart14, I will make new release

danielstuart14:
Any news about that?

PR #594, opened 2022-07-21
Enable extended advertising on android

danielstuart14:
RxAndroidBle 1.16.0 now supports enabling extended advertising (https://github.com/dariuszseweryn/RxAndroidBle/commit/4371b7832c1aaf328223f831d92b463dca039e85)
flutter_reactive_ble already supports extended advertising with iOS, so no changes are needed.

### DIFF
--- a/packages/reactive_ble_mobile/android/src/main/kotlin/com/signify/hue/flutterreactiveble/ble/ReactiveBleClient.kt
+++ b/packages/reactive_ble_mobile/android/src/main/kotlin/com/signify/hue/flutterreactiveble/ble/ReactiveBleClient.kt
@@ -75,6 +75,7 @@ open class ReactiveBleClient(private val context: Context) : BleClient {
         return rxBleClient.scanBleDevices(
             ScanSettings.Builder()
                 .setScanMode(scanMode.toScanSettings())
+                .setLegacy(false)
                 .setCallbackType(ScanSettings.CALLBACK_TYPE_ALL_MATCHES)
                 .setShouldCheckLocationServicesState(requireLocationServicesEnabled)
                 .build(),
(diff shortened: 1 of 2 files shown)


danielstuart14:
Solves https://github.com/PhilipsHue/flutter_reactive_ble/issues/571

packages/reactive_ble_mobile/android/src/main/kotlin/com/signify/hue/flutterreactiveble/ble/ReactiveBleClient.kt
```diff
@@ -75,6 +75,7 @@ open class ReactiveBleClient(private val context: Context) : BleClient {
         return rxBleClient.scanBleDevices(
             ScanSettings.Builder()
                 .setScanMode(scanMode.toScanSettings())
+                .setLegacy(false)
```
Taym95, line 78: why you set it to `false` here?

danielstuart14: Yes, it still works with 4.2 advertisings. setLegacy(false) enables the driver to receive extended advertisings (with more than 31 bytes), without removing support to old versions of BLE. (Google named it badly, "setOnlyLegacy" would be better, but anyhow...)
image

With this change, flutter_reactive_ble on android matches the iOS behaviour, returning both legacy and extended advertisings.

https://developer.android.com/reference/android/bluetooth/le/ScanSettings.Builder

Taym95: Thanks for the contribution!

Taym95 (approved):

PR #596, opened 2022-08-04
Read RSSI of connected device

### LINKED ISSUE #292: A way of getting RSSI of connected device
**Is your feature request related to a problem? (please describe)**
I do have a problem where i have to have signal strength fo my bluetooth device, the device stops advertising after being connected, so i am unable to read the RSSI.

it would be nice to have a possibility to listen to a stream of changing RSSI

**Describe alternatives you've considered**
i have considered scanning constantly but my device is not advertising after being connected.

**Additional context**
well thats pretty much all there is to it

TNorbury:
resolves #292 
Adds the ability to read the RSSI of a connected device. While the issue did ask for an RSSI stream I decided to go with just a Future instead. The reason being is the native BLE layers (to the best of my knowledge) don't have a mechanism for streaming RSSI, just reading the value once per call, as such I thought that it would be best to maintain that in the Flutter-side API.

I also updated the example app to include a Read RSSI button, mostly just as a proof of functionality, so it can be removed if need-be, you can see that demonstrated below:

### DIFF
--- a/packages/flutter_reactive_ble/lib/src/reactive_ble.dart
+++ b/packages/flutter_reactive_ble/lib/src/reactive_ble.dart
@@ -309,6 +309,12 @@ class FlutterReactiveBle {
       .clearGattCache(deviceId)
       .then((info) => info.dematerialize());
 
+  /// Reads the RSSI of the of the peripheral with the given device ID. 
+  /// The peripheral must be connected, otherwise a [PlatformException] will be
+  /// thrown
+  Future<int> readRssi(String deviceId) async =>
+      _blePlatform.readRssi(deviceId);
+
   /// Subscribes to updates from the characteristic specified.
   ///
   /// This stream terminates automatically when the device is disconnected.
--- a/packages/flutter_reactive_ble/lib/src/reactive_ble.dart
+++ b/packages/flutter_reactive_ble/lib/src/reactive_ble.dart
@@ -315,8 +315,10 @@ class FlutterReactiveBle {
   ///
   /// [rssiPeriod] controls the interval at which a new RSSI value will be
   /// returned. Defaults to 1 second.
-  Stream<int> streamRssi(String deviceId,
-          {Duration rssiPeriod = const Duration(seconds: 1)}) =>
+  Stream<int> streamRssi(
+    String deviceId, {
+    Duration rssiPeriod = const Duration(seconds: 1),
+  }) =>
       _blePlatform.streamRssi(deviceId, rssiPeriod);
 
   /// Subscribes to updates from the characteristic specified.
--- a/packages/flutter_reactive_ble/lib/src/reactive_ble.dart
+++ b/packages/flutter_reactive_ble/lib/src/reactive_ble.dart
@@ -309,11 +309,15 @@ class FlutterReactiveBle {
       .clearGattCache(deviceId)
       .then((info) => info.dematerialize());
 
-  /// Reads the RSSI of the of the peripheral with the given device ID. 
+  /// Streams the RSSI of the of the peripheral with the given device ID.
   /// The peripheral must be connected, otherwise a [PlatformException] will be
   /// thrown
-  Future<int> readRssi(String deviceId) async =>
-      _blePlatform.readRssi(deviceId);
+  ///
+  /// [rssiPeriod] controls the interval at which a new RSSI value will be
+  /// returned. Defaults to 1 second.
+  Stream<int> streamRssi(String deviceId,
+          {Duration rssiPeriod = const Duration(seconds: 1)}) =>
+      _blePlatform.streamRssi(deviceId, rssiPeriod);
 
   /// Subscribes to updates from the characteristic specified.
   ///
--- a/packages/reactive_ble_mobile/android/src/main/kotlin/com/signify/hue/flutterreactiveble/PluginController.kt
+++ b/packages/reactive_ble_mobile/android/src/main/kotlin/com/signify/hue/flutterreactiveble/PluginController.kt
@@ -275,7 +275,7 @@ class PluginController {
         bleClient.readRssi(args.deviceId)
                 .observeOn(AndroidSchedulers.mainThread())
                 .subscribe({ rssi ->
-                    val info = protoConverter.convertReadRssiResult(rssi);
+                    val info = protoConverter.convertReadRssiResult(rssi)
                     result.success(info.toByteArray())
                 }, { error ->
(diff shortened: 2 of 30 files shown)


packages/reactive_ble_mobile/lib/src/generated/bledata.pbenum.dart
```diff
@@ -0,0 +1,7 @@
+///
```
TNorbury, line 1: See other comments about generated side effects, for this file and all the other new ones in this directory

TNorbury:
@Taym95 I've had this running in production good w/o a problem. Lmk if there are any other changes you'd like to see.

FXschwartz:
@Taym95 Do you have an ETA on when you think you could review this? We would love to have this functionality.

Taym95:
> @Taym95 Do you have an ETA on when you think you could review this? We would love to have this functionality.

Sorry for long absence, I will review this week!

packages/reactive_ble_mobile/lib/src/reactive_ble_mobile_platform.dart
```diff
@@ -218,6 +218,16 @@ class ReactiveBleMobilePlatform extends ReactiveBlePlatform {
           )
           .then((data) => _protobufConverter.clearGattCacheResultFrom(data!));
 
+  @override
+  Future<int> readRssi(String deviceId) async => _bleMethodChannel
```
Taym95, line 222: Don't you think this should be a stream no a Future?

TNorbury: I went with a future to match what the underlying APIs did. But after giving it some thought I do agree that a stream would be more useful in terms actual use cases. 
I'll get that changed.

TNorbury: Done

packages/flutter_reactive_ble/lib/src/reactive_ble.dart
```diff
@@ -309,6 +309,16 @@ class FlutterReactiveBle {
...
       .then((info) => info.dematerialize());
 
+  /// Streams the RSSI of the of the peripheral with the given device ID.
+  /// The peripheral must be connected, otherwise a [PlatformException] will be
+  /// thrown
+  ///
+  /// [rssiPeriod] controls the interval at which a new RSSI value will be
+  /// returned. Defaults to 1 second.
+  Stream<int> streamRssi(String deviceId,
+          {Duration rssiPeriod = const Duration(seconds: 1)}) =>
```
Taym95, line 319: let's format the code a bit, maybe you can add `,`

Taym95:
We will need some test for this, let me know if you need some help!

packages/reactive_ble_mobile/android/src/main/kotlin/com/signify/hue/flutterreactiveble/PluginController.kt
```diff
@@ -267,4 +268,18 @@ class PluginController {
                 })
                 .discard()
     }
+
+    private fun readRssi(call: MethodCall, result: Result) {
+        val args = pb.ClearGattCacheRequest.parseFrom(call.arguments as ByteArray)
```
spkersten, line 273: I don't think `ClearGattCacheRequest` should be used for subscribing to RSSI updates.

packages/reactive_ble_mobile/ios/Classes/Plugin/PluginController.swift
```diff
@@ -523,6 +523,50 @@ final class PluginController {
...
+                        }
+                    )
+                }
+            )
+        } catch let error {
+            completion(.failure(makeFlutterError(error: error)))
+        }
+    }
+
+    // takes an error and converts it into a Flutter error
```
spkersten, line 562: This comment can be omitted (the function name gives the same information)

packages/reactive_ble_mobile/ios/Classes/ReactiveBle/Central.swift
```diff
@@ -281,6 +288,26 @@ final class Central {
...
+        let peripheral = try resolve(connected: peripheralId)
+
+        // register task, completion goes straight to the caller
+        readRssiRegistry.registerTask(
+            key: peripheralId,
+            params: .init(),
+            completion: completion
+        )
+
+        // update task to kick things off
```
spkersten, line 302: These two comments can be removed as well IMHO, as they don't give information that isn't already clear.

packages/reactive_ble_mobile/lib/src/reactive_ble_mobile_platform.dart
```diff
@@ -218,6 +218,20 @@ class ReactiveBleMobilePlatform extends ReactiveBlePlatform {
           )
           .then((data) => _protobufConverter.clearGattCacheResultFrom(data!));
 
+  @override
+  Stream<int> streamRssi(String deviceId, Duration rssiPeriod) =>
+      Stream.periodic(
+        rssiPeriod,
+        (computationCount) => _bleMethodChannel
```
spkersten, line 225:
```suggestion
        (_) => _bleMethodChannel
```

packages/reactive_ble_mobile/lib/src/reactive_ble_mobile_platform.dart
spkersten: Sorry open this discussion as you already changed it once, but is a Stream really the best option here? On both Android and iOS, the API is to read the RSSI once. Why would our library add a periodic Stream on top of that? It is more flexible for the user to have the option to read it once, while they can easily write Stream.periodic themselves if they need to.

smofe:
Is there any update regarding this? This would be super useful for us to have for a project.

qq1023308356 (changes requested):
I really need this function

ky1vstar:
Seriously? It's been six months already! This is such a simple and useful feature, but you are wasting collaborator's time on removing some comment?

raphael-bmec-co:
Very keen to see this implemented. Can we get details on the failed check and assist with resolution?

azhansy:
Is there any update regarding this?

arnoschn:
Second that! Would be a very helpful feature.

spkersten:
Since this PR seems to be abandoned, I've taken over the work, added some tests and addressed the review comments here: https://github.com/PhilipsHue/flutter_reactive_ble/pull/796

PR #599, opened 2022-08-11
Federated implementations

trik:
Following the discussion in #544, i'm trying to publish a web implementation of this plugin but i'm facing a couple of issues:

- ReactiveBlePlatform class exposes the set instance method to enable other platform-specific implementations but the instance is currently always re-set using ReactiveBleMobilePlatformFactory https://github.com/PhilipsHue/flutter_reactive_ble/blob/master/packages/flutter_reactive_ble/lib/src/reactive_ble.dart#L107

- platformIsAndroid variable is currently set using Platform.isAndroid, which rises an exeception in non-native environments

### DIFF
--- a/packages/flutter_reactive_ble/lib/src/reactive_ble.dart
+++ b/packages/flutter_reactive_ble/lib/src/reactive_ble.dart
@@ -1,6 +1,7 @@
 import 'dart:async';
 import 'dart:io';
 
+import 'package:flutter/foundation.dart' show kIsWeb;
 import 'package:flutter_reactive_ble/src/connected_device_operation.dart';
 import 'package:flutter_reactive_ble/src/debug_logger.dart';
 import 'package:flutter_reactive_ble/src/device_connector.dart';
@@ -116,7 +117,7 @@ class FlutterReactiveBle {
       );
       _deviceScanner = DeviceScannerImpl(
         blePlatform: _blePlatform,
-        platformIsAndroid: () => Platform.isAndroid,
+        platformIsAndroid: () => !kIsWeb && Platform.isAndroid,
         delayAfterScanCompletion: Future<void>.delayed(
           const Duration(milliseconds: 300),
         ),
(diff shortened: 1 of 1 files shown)


packages/flutter_reactive_ble/lib/src/reactive_ble.dart
```diff
@@ -104,8 +105,11 @@ class FlutterReactiveBle {
         print,
       );
 
-      ReactiveBlePlatform.instance =
-          const ReactiveBleMobilePlatformFactory().create();
+      // ignore: unnecessary_null_comparison, prefer_conditional_assignment
+      if (ReactiveBlePlatform.instance == null) {
+        ReactiveBlePlatform.instance =
```
Taym95, line 110: ` ReactiveBlePlatform.instance ??= const ReactiveBleMobilePlatformFactory().create();`

Taym95 (changes requested):

packages/flutter_reactive_ble/lib/src/reactive_ble.dart
```diff
@@ -116,7 +120,7 @@ class FlutterReactiveBle {
       );
       _deviceScanner = DeviceScannerImpl(
         blePlatform: _blePlatform,
-        platformIsAndroid: () => Platform.isAndroid,
+        platformIsAndroid: () => !kIsWeb && Platform.isAndroid,
```
Taym95, line 123: :+1:

Taym95 (approved):

trik:
@Taym95 i think this code needs to be refactored. it's working good with a custom implementation but it's throwing a LateInitializationError when using the default one (reactive_ble_mobile).

```dart
try {
  ReactiveBlePlatform.instance;
} on LateInitializationError {
  ReactiveBlePlatform.instance = const ReactiveBleMobilePlatformFactory().create();
}
```

Issue #211, opened 2021-02-02
First quick notifications seem to be ignored


Klievan:
**Describe the bug**
I have a BLE device that instantly starts sending notifications once they are turned on. It seems that the first 1-2 notifications seem to be not sent to the stream I am listening to. Also, if I only send those 2 notifications without sending any other, none are received.

**To Reproduce**
1. Connect to a BLE device that instantly sends notifications when written to the CCCD to enable notifications
2. Subscribe to the characteristic and print out all incoming notifications
3. Observe that the first data does not seem to be received 

**Expected behavior**
I expect the first notifications to be also received.

- [x] When using nRFConnect or flutter_blue I am able to receive all notifications.

**Smartphone / tablet**
 - Device: OnePlus 5T
 - OS:  Android 10
 - Package version: ^2.7.2

**Peripheral device**
- Vendor, model: Both the Silabs BGM121 and nRF52832 appear to have the same behavior 
- Does it run a custom firmware / software: yes

remonh87:
This is a very specific usecase and I really need your help in triaging this. Can you tryout the same usecase but then set the loglevel to verbose?  you can use the setter `logLevel` on the `FlutterReactiveBle` client for this

Klievan:
```
I/flutter (10367): REACTIVE_BLE: Start subscribing to notifications for QualifiedCharacteristic(characteristicId: f3641404-00b0-4240-ba50-05ca45bf8abc, serviceId: f3641400-00b0-4240-ba50-05ca45bf8abc, deviceId: 00:0B:57:BE:7E:32)
D/BluetoothGatt(10367): setCharacteristicNotification() - uuid: f3641404-00b0-4240-ba50-05ca45bf8abc enable: true

```
This appears but no notifications are received (even though 2 already have been sent). The ones that arrive about 200ms later do get registered though. The use case is a file being streamed through notifications once they are enabled, but the first notifications seem to get accepted by the phone but do not get passed onto the app.

remonh87:
hmm I had hoped to see more logging then this 🤔 . If you want I can creste a triage branch with verbose logging but I may be able to do this the coming days. Problem is that I do not have the exact same peripheral which makes it difficult to check.

Klievan:
I can be actively debugging the issue with you if needed, or if you happen to have an nRF52 devkit of some sorts I could send you a firmware example. I've further tested this on an iOS device and I get the following assert:
```
[...]

PR #685, opened 2023-02-07
Enable quick-setup by default

Klievan:
Fixes https://github.com/PhilipsHue/flutter_reactive_ble/issues/211

Setting the NotificationSetupMode to QUICK_SETUP makes it so the notification-observable stream gets opened BEFORE writing the descriptor on the peripheral. This makes it so that if the peripheral starts sending notifications instantly upon subscribing, no data is lost. Otherwise, the couple of milliseconds between writing the notification-descriptor and propagating the observable stream to the Flutter app, notifications can be received by the phone that are missed by the app.

Other BLE-libraries seem to be also using AndroidRxBle but did not have this issue, so I can only assume they were using the same mode.


Klievan:
Not sure why the merge is failing checks. Can anyone access the build log to let me know what needs fixing?

PR #693, opened 2023-02-18
macos version

alanswx:
Unfortunately, I am not sure of the best way to merge this together. The macos has a couple of define checks so that it is the same code on ios and macos. Unfortunatley, just doing a symlink for the Classes directory doesn't work. It won't compile correctly. Here are some comments on different ways to reuse the same code:

https://github.com/flutter/flutter/issues/52285#issuecomment-596914611


Taym95:
@alanswx thank you very much for contribution, we agreed with our team to keep this repo for IOS/Android, and we are open to endorsing plugins for other platforms and link in readme.

PR #703, opened 2023-03-02
Remove dead code from reactive_ble_platform_test

srawlins:
`DEAD_CODE` will soon be a warning which will start to fail `dart analyze` (without `--fatal-warnings=false`).

This change will prevent that warning from being reported.


srawlins:
Gentle ping.

Taym95 (approved):

PR #721, opened 2023-04-13
Remove asyncness from flutter_reactive_ble_exampel:ble_device_connector.dart:connect

msandstedt:
This doesn't need to be an async future because it never awaits. Therefore, for clarity, remove its asyncness.


Taym95:
Thanks for contribution, I will close this PR since it not active and feel free to open it again.

PR #723, opened 2023-04-22
Make Plugin Controller not static so it creates new instance for each Flutter Engine

DavidDomkar:
I encountered a bug with this package when using together with flutter_foreground_task plugin.

flutter_foreground_task plugin creates a separate Flutter Engine in android foreground task to perform some work. This engine also initializes flutter_reactive_ble plugin.

However, because the PluginController instance is static, it would deinitialize the plugin when Flutter Engine that the app uses is destroyed (on app force close) which leaves the plugin deinitialized for the potentionally running Flutter Engine in the foreground task.

This PR ensures that each Flutter Engine has its own instance of the PluginController which deintializes only if the given engine is destroyed. This allows for plugin usage within background contexts like the foreground task on android.

### DIFF
--- a/packages/reactive_ble_mobile/android/src/main/kotlin/com/signify/hue/flutterreactiveble/ReactiveBlePlugin.kt
+++ b/packages/reactive_ble_mobile/android/src/main/kotlin/com/signify/hue/flutterreactiveble/ReactiveBlePlugin.kt
@@ -1,40 +1,22 @@
 package com.signify.hue.flutterreactiveble
 
-import android.content.Context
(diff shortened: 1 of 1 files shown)


dominicmh:
This fix is important since the underlying bug turns the plugin useless in any project that creates – or has plugins that create – a new FlutterEngine.
An additional issue remains: Whenever a FlutterEngine is destroyed the plugin will also disconnect from all connections or stop scanning. E.g. if a background task is finished or in hybrid apps that have some parts of the UI native and some parts in Flutter.

Issue #43, opened 2020-03-11
No way of knowing if peripheral is connectable


juliansteenbakker:
**Is your feature request related to a problem? (please describe)**
Right now it's not possible to check if the found peripheral is connectable. There are devices that advertise without being connectable, so it would be good to filter out those and only allow connectable devices.

**Describe the solution you'd like**
For iOS, CBAdvertisementDataIsConnectable lets you know wether a peripheral is connectable. It would be great if this gets passed through.
However, since this plugin relies on RxAndroidBle, it is currently not possible to check if a peripheral is connectable on Android. See ticket https://github.com/Polidea/RxAndroidBle/issues/668

**Describe alternatives you've considered**
x

**Additional context**
x

remonh87:
Thanks for reporting this I think this will be a good addition.

RtgrV:
@Taym95 we should be able to close this issue after releasing a new version.

PR #750, opened 2023-06-27
Add connectability from discovery data

RtgrV:
Issue report: https://github.com/PhilipsHue/flutter_reactive_ble/issues/43

- Upgraded packages and ran codegen
- Added Connectable field in discovery data on Android and iOS
- Updated iOS project files

| Android | iOS | iOS detail |
| --- | --- | --- |
| image | Image from iOS | Image from iOS (1) |


example/lib/src/ui/device_detail/device_detail_screen.dart
```diff
@@ -39,7 +39,7 @@ class _DeviceDetail extends StatelessWidget {
           length: 2,
           child: Scaffold(
             appBar: AppBar(
-              title: Text(device.name),
+              title: Text(device.name.isNotEmpty ? device.name : "Undefined"),
```
werediver, line 42: Would "Unnamed" make more sense?

example/lib/src/ui/device_list.dart
```diff
@@ -153,16 +153,26 @@ class _DeviceListState extends State<_DeviceList> {
                   ...widget.scannerState.discoveredDevices
                       .map(
                         (device) => ListTile(
-                          title: Text(device.name),
-                          subtitle: Text("${device.id}\nRSSI: ${device.rssi}"),
+                          title: Text(
+                            device.name.isNotEmpty ? device.name : "Undefined",
```
werediver, line 157: Same, consider "Unnamed" instead.

packages/reactive_ble_mobile/protos/bledata.proto
```diff
@@ -13,9 +13,10 @@ message DeviceScanInfo {
     string name = 2;
     GenericFailure failure = 3;
     repeated ServiceDataEntry serviceData = 4;
-    bytes manufacturerData = 6;
-    repeated Uuid serviceUuids = 7;
+    bytes manufacturerData = 7;
+    repeated Uuid serviceUuids = 8;
     int32 rssi = 5;
+    IsConnectable isConnectable = 6;
```
werediver, line 19: Not sure if we need to maintain compatibility with the older versions here, but, either way, reshuffling numbers like this makes no sense. The order is both weird _and_ not compatible with the previous version.

While it could make sense to stay backwards compatible, in theory, I don't think it is important for this project practically.

Please, choose one or the other approach and either don't change the old numbers, or put them all in order.

werediver (approved):
Please, check my remarks. Because they are not critical, I approve any way.

Thanks for contributing.

Taym95 (approved):

Issue #641, opened 2022-12-06
Provide separate access to device name in Generic Access Profile and Advertisement Data.


abarresi-abamis:
**Is your feature request related to a problem? (please describe)**
This library was updated to address #84 and #105, but since these fixes use the advertisement data local name (if provided), developers no longer have separate access to the device name defined in the Generic Access Profile if they choose to also use the advertisement data local name.

This can be worked around on Android since you can access the GAP service to read the device name characteristic, but on iOS, the GAP service is blocked by Apple and the only way to retrieve it is via Apple's provided peripheral.name, which this library only uses if the advertisement data local name is not provided. It seems this functionality was changed in this library due to Apple's caching of the GAP device name, and them not providing developers control over clearing this cache, but not providing it via this library seems incomplete and excludes some valid use cases.

**Describe the solution you'd like**
Provide access to both the local name defined in the advertisement data, and the device name defined in the Generic Access Profile as separate properties.

Issue #628, opened 2022-11-10
add createBond / Bonding status 


FaroukBoussarsar:
when connect to device using android double pairing request is received.
so as a work around if it is possible to bond from the side of the app it will solve this issue

https://developer.android.com/reference/android/bluetooth/BluetoothDevice#createBond()

this feature also helps detect if bounding is done or not so it minimise handling from our side

vincent-iQontrol:
Indeed, it would be nice to implement some sort of bonding helper: https://gist.github.com/dglozano/9b0ce38a558eeca16137909bd368698c
We currently facing problems on android devices with "just works" implementation when all services are hidden before bonding.

Issue #507, opened 2022-01-27
Android pairing request popup appears twice


BradKwon:
**Describe the bug**
When a bluetooth peripheral, which has a Nordic chip (I am not sure this happens only with this chip though), is connected to some Android phones, the Android native pairing request popup appears twice. The first popup does nothing.

**To Reproduce**
1. Run example Android app
2. Connect the bluetooth peripheral, possibly with a Nordic chip.
3. Check if Android native pairing request popup appears.
4. Click "pair" button on it.
5. Observe the 2nd pairing popup appears.

**Expected behavior**
The Android native pairing popup appears only once.

- [ ] I tried doing the same with a general BLE scanner application (e.g. [nRF Connect](https://duckduckgo.com/?q=nRF+connect)) and it exhibits the expected behavior as described above

**Smartphone / tablet**
 - Device: Google Pixel 4a
 - OS: Android 12
 - Package version: 5.0.1

**Peripheral device**
- Vendor, model: Custom with Nordic chip
- Does it run a custom firmware / software: no

**Additional context**
This also happens with the general BLE scanner app however, I am wondering if there are any workaround for it.

PieterAelse:
Can you reproduce the issue with any other BLE device than the custom Nordic chip? As we're not seeing this issue in the field, it might have to do with your device (even though the nRF Connect app doesn't show it).

And to be precies; what "general BLE scanner app" do you mean?

BradKwon:
@PieterAelse I have tested it again with other BLE mouse. I think this seems not a Nordic chip. I was not able to find which chip this has. The pairing request popup appears twice as well with it. 

I noticed that it appears only once on my two Android 9 devices, but on the Android 12 device, Google Pixel 4a.

**_My Android 9 devices:_**
Motorola X4
LG G2 w/ Lineage OS

I mean the nRF Connect app by "general BLE scanner app". Sorry for the confusion. I just used the term in the template.

Based on my further test, the issue only happens on Android 12. Unfortunately, I have only Google Pixel 4a so cannot test it on other Android 12 devices. And the nRF Connect app also shows it twice so it looks like a general issue.

It would be nice someone else report this or can test it on other Android 12 devices. And it would be nice if there are any workaround.

GaelleJoubert:
I have the same thing happening to me. 
On a Pixel 4A5G with Android 12. 
If I don't accept it twice, it's not working afterwards. 
I try my code on two other phone :
[...]

Issue #272, opened 2021-05-15
Support for Companion device pairing?


lwinged:
Hello,

I'd like to know if you will integrate this feature to avoid ask the user the location permission ?

https://developer.android.com/guide/topics/connectivity/companion-device-pairing

My app needs to scan devices but I don't need/want the  location permission. (ACCESS_FINE_LOCATION)

My app targets min sdk Android 7

I saw that the scan method has a parameter : requireLocationServicesEnabled

With this parameter can I scan without ask the user for location permission?

Perhaps do you have any solutions about that?

Do I need implement this feature in my app for Android 8+ or the companion devices pairing can be a feature/ an evolution of ble library ?

Thank you very much.

remonh87:
Hi 
I understand your request and I agree that the choice of Google to tie the location permission to ble scanning is not ideal and can raise privacy concerts.  

I think it can be a good addition to our library in the future given that the underlying [Android library](https://github.com/Polidea/RxAndroidBle/issues/721) supports it.  Also since it works on the background we need to figure out how to nicely integrate it.  @werediver  do you know if iOS has something that looks like this? 

 Keep in mind that companion device pairing is a totally different user experience than "normal scanning/connecting/pairing" . Users are offered a dialog of devices in the vicinity and will be provide a dialog to link it to the app. 

The `requireLocationServicesEnabled ` parameter just makes sure that the stream throws an exception that you can handle in case the location is not enabled. Reason for this is that on some Android devices (especially Nexus) ble scanning does not yield results when the users has location service off. 

On the short term I have other priorities but feel free to contribute to this library. Also I am willing to help out.

PR #760, opened 2023-07-04
Adds support for Android companion pairing, device bonding, iOS device name (via GAP)

ened:
This addresses #272, #628, #641 in the same PR. 

For a project, I needed multiple additions, so the code is quite mixed. 
Let's look at this PR holistically first. (I needed both sets of functionality for a project).

Whole set of changes:
 - add a method to retrieve the device name via the `CBPeripheral.name` method, partially addresses #641 
 - add bonding configuration for new connections. this addresses #507 as well on my side, but needs further testing
 - integrates with the (foreground only for now) CompanionDeviceManager system on Android (see https://developer.android.com/guide/topics/connectivity/companion-device-pairing for details - I was able to drop a permission - `scan` - from my App, which is great)


remonh87 (changes requested):
Thanks for the PR I think it will be a great addition to the library. I will look into the implementation further this weekend but had now half hour time to overlook the structure. I am a bit concerned about adding the bond mode to connect method. Maybe it is better we add a specific method for companion pairing.

packages/flutter_reactive_ble/lib/src/reactive_ble.dart
```diff
@@ -256,16 +270,31 @@ class FlutterReactiveBle {
     required String id,
     Map<Uuid, List<Uuid>>? servicesWithCharacteristicsToDiscover,
     Duration? connectionTimeout,
+    BondingMode? bondingMode,
```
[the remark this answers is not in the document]
ened: I think this makes sense. So _bond_ first, then another call to _connect_.
Companion is in a different context here, (sorry for the confusion), so I think we can call it `establishBonding`.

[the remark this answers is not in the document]
remonh87: ah yes lets call it establishBonding. We should throw an exception on iOS since this is not supported for that platform.

packages/reactive_ble_mobile/android/src/main/kotlin/com/signify/hue/flutterreactiveble/PluginController.kt
```diff
@@ -132,84 +157,113 @@ class PluginController {
...
-        executeWriteAndPropagateResultToChannel(call, result, com.signify.hue.flutterreactiveble.ble.BleClient::writeCharacteristicWithoutResponse)
+        executeWriteAndPropagateResultToChannel(
+            call,
+            result,
+            com.signify.hue.flutterreactiveble.ble.BleClient::writeCharacteristicWithoutResponse
+        )
     }
 
     private fun executeWriteAndPropagateResultToChannel(
-            call: MethodCall,
```
remonh87, line 182: this big change set makes me think we should enforce standard formatting rules so everyone formats the same

packages/reactive_ble_mobile/android/src/main/kotlin/com/signify/hue/flutterreactiveble/PluginController.kt
```diff
@@ -227,44 +281,87 @@ class PluginController {
...
+            .discard()
+    }
+
+    fun setActivity(activity: Activity?) {
+        companionHandler.setActivity(activity)
+    }
+
+    override fun onActivityResult(requestCode: Int, resultCode: Int, data: Intent?): Boolean {
+        if (requestCode == CompanionHandler.SELECT_DEVICE_REQUEST_CODE && resultCode == Activity.RESULT_OK) {
+            Log.d(TAG, "received result from device selection activity, parsing via companion handler")
```
remonh87, line 352: let's remove the debug logs 

- [ ] remove debug logs

packages/reactive_ble_mobile/android/src/main/kotlin/com/signify/hue/flutterreactiveble/ble/BondingManager.kt
```diff
@@ -0,0 +1,104 @@
...
+import io.reactivex.Completable
+import io.reactivex.disposables.Disposables
+
+/**
+ * Created by clement on 2017-08-24.
+ */
+
+class BondingFailedException : RuntimeException()
+
+object BondingManager {
```
remonh87, line 22: can you make this a class ?  I prefer  to use this so it is easier for mocking and testing than an object.  Also let's add a unites for this function

- [ ]  add unites.

packages/reactive_ble_mobile/android/src/main/kotlin/com/signify/hue/flutterreactiveble/channelhandlers/CompanionHandler.kt
```diff
@@ -0,0 +1,140 @@
...
+        }
+    }
+
+    fun setActivity(activity: Activity?) {
+        this.activity = WeakReference(activity);
+    }
+
+    @RequiresApi(Build.VERSION_CODES.O)
+    fun onActivityResult(data: Intent?): ScanResult? {
+        Log.d(TAG, "onActivityResult: $data")
```
remonh87, line 119: let's also remove here the debug logging

packages/reactive_ble_mobile/android/src/main/kotlin/com/signify/hue/flutterreactiveble/model/BondingMode.kt
```diff
@@ -0,0 +1,9 @@
+package com.signify.hue.flutterreactiveble.model
+
+import android.content.Context
+
+enum class BondingMode(val code : Int) {
```
remonh87, line 5: I think we can get rid of this right when we go for a separate call to create bond?

packages/reactive_ble_mobile/ios/Classes/Plugin/PluginController.swift
```diff
@@ -258,6 +258,31 @@ final class PluginController {
         }
     }
 
+    func retrieveDeviceName(name: String, args: GetDeviceNameRequest, completion: @escaping PlatformMethodCompletionHandler) {
```
remonh87, line 261: @werediver  can look into the iOS implementation?

packages/reactive_ble_mobile/protos/bledata.proto
```diff
@@ -2,6 +2,16 @@ syntax = "proto3";
 option java_package = "com.signify.hue.flutterreactiveble";
 option java_outer_classname = "ProtobufModel";
 
+message LaunchCompanionRequest {
+    string pattern = 1;
```
remonh87, line 6: I think we should call it everywhere deviceNamePattern what do you think? Else people can be confused what it is.

- [ ]  Change to `deviceNamePattern `

packages/reactive_ble_platform_interface/lib/src/model/association_info.dart
```diff
@@ -0,0 +1,16 @@
+import 'package:functional_data/functional_data.dart';
+import 'package:meta/meta.dart';
+
+part 'association_info.g.dart';
+
+//ignore_for_file: annotate_overrides
+
+@immutable
+@FunctionalData()
+class AssociationInfo extends $AssociationInfo {
```
remonh87, line 10: is DeviceAssociationInfo a better name? Then we can call the property `macAddress`. Also let's add docs to the class

packages/reactive_ble_platform_interface/lib/src/model/bonding_mode.dart
```diff
@@ -0,0 +1,13 @@
+/// Specify the bonding mode for the device.
+enum BondingMode {
```
remonh87, line 2: Like discussed above I think if we have a separate method for establish bonding we do not need this.

packages/reactive_ble_platform_interface/lib/src/reactive_ble_platform_interface.dart
```diff
@@ -68,6 +68,16 @@ abstract class ReactiveBlePlatform extends PlatformInterface {
     throw UnimplementedError('deInitialize() has not been implemented.');
   }
 
+  /// Launch Companion workflow for connecting a Ble device (Android only)
+  Future<AssociationInfo?> launchCompanionWorkflow({
+    required String pattern,
```
remonh87, line 73: deviceNamePattern

remonh87 (changes requested):
Now the full review: again thank you very much for implementing this I think it is close to be integrated. Please check out my remarks. Also some generic things:
- Add unit tests for the compianonflow on dart
- Add unit tests for the protobuf conversion methods related to device name and companion flow
- Document the public methods Especially with `createBond` it is important that we emphasise that this will only work on android. 

Reach out if you need help

packages/reactive_ble_mobile/android/src/main/kotlin/com/signify/hue/flutterreactiveble/PluginController.kt
ened: Let's agree on a set (I think dartfmt, ktlint, swiftlint), th en add as part of CI.
Happy to have a stab. :-)

packages/reactive_ble_mobile/android/src/main/kotlin/com/signify/hue/flutterreactiveble/model/BondingMode.kt
ened: I removed `BondingMode`, but introduced a new `BondingStatus` enum. The enum reflects the Android states (none, bonding, bonded) and could be used in the future for an observing component as well. Point is there's a need to know the outcome of the bonding (and "failed"/not bonded may not always be a problem for the calling app). Therefore we return the actual value here.

packages/reactive_ble_mobile/ios/Classes/Plugin/PluginController.swift
```diff
@@ -258,6 +258,31 @@ final class PluginController {
...
+
+        do {
+            let name: String? = try central.retrievePeripheralName(for: deviceID)
+            let message = DeviceNameInfo.with {
+                $0.id = deviceID.uuidString
+                $0.deviceName = name ?? ""
+            }
+            completion(.success(message))
+        } catch {
+            completion(.success(nil))
```
werediver, line 282: This branch doesn't look like success to me.

packages/reactive_ble_mobile/ios/Classes/ReactiveBle/Central.swift
```diff
@@ -167,6 +167,14 @@ final class Central {
...
             .forEach(centralManager.cancelPeripheralConnection)
     }
+    
+    func retrievePeripheralName(
+        for peripheralID: PeripheralID
+    ) throws -> String? {
+        let peripheral = try resolve(known: peripheralID)
+        
+        return peripheral.name
+    }
```
werediver, line 177: I doubt the value of this addition. Doesn't look much better than `try c.resolve(known: peripheralID)?.name`.

You're not suggesting to create separate methods here for every property of `CBPeripheral` that we may need to access, are you?

werediver: Well, it seems the original design-philosophy here was not to expose CoreBluetooth types directly, which justifies your addition (`resolve(known:)` and co. are private).

If you want to stream the updates, on of the `CB-` delegates should be involved.

ened: So for this iteration I would suggest not to add the streaming capability yet. 

If the structure of the addition is ok, then we can probably agree on this approach for now?

ened:
Rebased onto master and worked through the review. `establishBonding` is its own method now (throws `UnsupportedError` off Android), the debug logs are gone, and `BondingManager` is a class with tests. Renamed to `deviceNamePattern` throughout, `DeviceAssociationInfo`/`macAddress` with docs, and kept `BondingStatus` (no `BondingMode`). Added tests for the companion flow and the proto conversions, and fixed the two iOS bits — `retrievePeripheralName` and that `PluginController` branch which wasn't actually a success path.

PR #763, opened 2023-07-05
Migrate to latest Android SDK, AGP and protobuf plugin

ened:
- Bumps the AGP and Gradle versions, so that modern Android Studio can compile happily.
- Bumps the compile and targetSDK to 33, so that new APIs are available (needed for #760)
- Update the protobuf plugin, as this would cause issues on modern build environments (like this one, but also from developer reports, see #761)

fixes #630 
fixes #761 

closes #754

PR #767, opened 2023-07-10
Adds namespace to `build.gradle` to support Gradle 8

Rexios80:


Taym95:
in progress part of #763

PR #776, opened 2023-07-26
Support multiple services or characteristics with the same id

### LINKED ISSUE #478: Support for multiple characteristics with the same Uuid
Some BLE Profiles allow for having multiple characteristics within a service that have the same Uuid and are uniquely identified by the characteristic descriptor (e.g. Automation IO).

Is it possible to support multiple characteristics with the same Uuid within a single service?

For example, if the `List<DiscoveredCharacteristic>` in the `DiscoveredService` class had an additional unique field so that the list could contain multiple characteristics with the same Uuid.

I understand that the second 'u' in 'Uuid' already implies that this should uniquely identify the characteristic :)
But the Automation IO Service specification states:

_"If a device has more than one instance of the Analog characteristic, each characteristic shall include a Characteristic Presentation Format descriptor that has a namespace / description value that is unique for that instance of the Analog characteristic... Description values from 0x0001 and upwards shall be used to uniquely identify each Analog characteristic."_

[https://www.bluetooth.com/specifications/specs/automation-io-service-1-0/](url)

### LINKED ISSUE #509: writeCharacteristic does not respect service UUID in Android implementation
**Describe the bug**
When interacting with a BLE device that has the same characteristic UUIDs under different service UUIDs, the write characteristic commands do not ensure writing to the correct characteristic, even though the supplied parameter of type QualifiedCharacteristic contains the correct combination of service UUID and characteristic UUID.

**Expected behavior**
Addressing a characteristic with a unique combination of characteristic UUID and service UUID should be possible.

**Additional context**
https://github.com/PhilipsHue/flutter_reactive_ble/blob/master/packages/reactive_ble_mobile/android/src/main/kotlin/com/signify/hue/flutterreactiveble/ble/ReactiveBleClient.kt

For example, lines 176 - 186:
```
override fun writeCharacteristicWithResponse(
        deviceId: String,
        characteristic: UUID,
        value: ByteArray
    ): Single<CharOperationResult> =
        executeWriteOperation(
            deviceId,
            characteristic,
            value,
            RxBleConnection::writeCharWithResponse
        )
```

As the function only takes the charactistic UUID but not the service UUID, it *cannot* ensure writing to the correct characteristic in cases where the same characteristic UUID is present in multiple services. (The same holds for other functions in this file.)

On the other hand, the iOS implementation makes sure to respect the service:
https://github.com/PhilipsHue/flutter_reactive_ble/blob/master/packages/reactive_ble_mobile/ios/Classes/ReactiveBle/Tasks/CharacteristicWrite/CharacteristicWriteTaskController.swift

```
guard
    let service = peripheral.services?.first(where: { $0.uuid == task.key.serviceID }),
    let characteristic = service.characteristics?.first(where: { $0.uuid == task.key.id }),
    characteristic.properties.contains(.write)
```

spkersten:
Fixes #478
Fixes #509

The change is mostly backwards compatible. The exception to that is that if you try to use the existing read/write/subscribe methods for a characteristic ids for which there is no single unique instance, an exception is thrown. Before, it would read/write/subscribe the first matching characteristic.


packages/reactive_ble_mobile/ios/Classes/ReactiveBle/Central.swift
```diff
@@ -82,17 +84,29 @@ final class Central {
                 )
             },
             onCharacteristicNotificationStateUpdate: papply(weak: self) { central, characteristic, error in
+                guard let q = try? QualifiedCharacteristic(characteristic)
+                else {
+                    print("Swift:X1");
```
spkersten, line 89: Remove

packages/reactive_ble_mobile/ios/Classes/ReactiveBle/Central.swift
```diff
@@ -82,17 +84,29 @@ final class Central {
...
                 )
             },
             onCharacteristicValueUpdate: papply(weak: self) { central, characteristic, error in
-                onCharacteristicValueUpdate(central, QualifiedCharacteristic(characteristic), characteristic.value, error)
+                onCharacteristicValueUpdate(central, try! QualifiedCharacteristic(characteristic), characteristic.value, error)
             },
             onCharacteristicValueWrite: papply(weak: self) { central, characteristic, error in
+                guard let q = try? QualifiedCharacteristic(characteristic)
+                else {
+                    print("Swift:X2");
```
spkersten, line 104: Remove

packages/reactive_ble_mobile/android/src/main/kotlin/com/signify/hue/flutterreactiveble/ble/ReactiveBleClient.kt
```diff
@@ -152,24 +154,31 @@ open class ReactiveBleClient(private val context: Context) : BleClient {
         }.firstOrError()
     }
 
+    @SuppressLint("MissingPermission")
```
Taym95, line 157: I recommend against adding `@SuppressLint("MissingPermission")` to the library in order to suppress lint warnings. Instead, we have provided clear instructions in the README about the required permissions for the library's proper functioning. It's important to maintain a transparent approach to permissions and ensure that developers integrating the library are aware of and handle the necessary permissions appropriately.

packages/flutter_reactive_ble/lib/src/reactive_ble.dart
```diff
@@ -290,54 +294,256 @@ class FlutterReactiveBle {
...
+  /// Note: On Android, this method performs service discovery if one was done yet after connecting the device.
+  Future<List<Service>> getDiscoveredServices(String deviceId) async {
+    final discoveredServices = await _connectedDeviceOperator.getDiscoverServices(deviceId);
+
+    final services = _services[deviceId] ?? [];
+
+    for (final discoveredService in discoveredServices) {
+      final service = services.firstWhere(
+        (service) =>
+            service.id == discoveredService.serviceId && service._instanceId == discoveredService.serviceInstanceId,
```
Taym95, line 332: Maybe we can reuse this part in line 371:

```
bool _isMatchingService(Service service, DiscoveredService discoveredService) {
  return service.id == discoveredService.serviceId && service._instanceId == discoveredService.serviceInstanceId;
}
``

File: packages/flutter_reactive_ble/lib/src/reactive_ble.dart [hunk 5]
```diff
@@ -290,54 +294,256 @@ class FlutterReactiveBle {
...
+          );
+          services.add(newService);
+          return newService;
+        },
+      );
+
+      for (final discoveredCharacteristic in discoveredService.characteristics) {
+        if (!service._characteristics.any((char) =>
+            char.id == discoveredCharacteristic.characteristicId &&
+            char._instanceId == discoveredCharacteristic.characteristicInstanceId)) {
```
Taym95, line 347: Maybe we can reuse this also: 363

```
bool _isMatchingCharacteristic(Characteristic char, DiscoveredCharacteristic discoveredCharacteristic) {
  return char.id == discoveredCharacteristic.characteristicId && char._instanceId == discoveredCharacteristic.characteristicInstanceId;
}
```

Taym95 (changes requested):

packages/flutter_reactive_ble/lib/src/reactive_ble.dart
```diff
@@ -290,54 +294,256 @@ class FlutterReactiveBle {
...
+
+  /// Services can be discovered by:
+  /// - specify `servicesWithCharacteristicsToDiscover` in [connectToDevice] or [connectToAdvertisingDevice]. Or,
+  /// - by calling [discoverAllServices]
+  ///
+  /// Discovered services and charactersitcs remain valid for as long as the device stays connected. After reconnecting
+  /// to a device, [getDiscoveredServices] must be called again to get updated services and characteristics.
+  ///
+  /// Note: On Android, this method performs service discovery if one was done yet after connecting the device.
+  Future<List<Service>> getDiscoveredServices(String deviceId) async {
```
Taym95, line 324: I guess we should write test for this function ?

packages/reactive_ble_mobile/android/src/main/kotlin/com/signify/hue/flutterreactiveble/ble/ReactiveBleClient.kt
spkersten: How does adding a `SuppressLint` conflict with "maintain[ing] a transparent approach to permissions and ensur[ing] that developers integrating the library are aware of and handle the necessary permissions appropriately"?

packages/flutter_reactive_ble/lib/src/reactive_ble.dart
```diff
@@ -290,54 +294,258 @@ class FlutterReactiveBle {
...
+        }
+      }
+    }
+    for (final service in services.toList()) {
+      if (!discoveredServices.any((discoveredService) => _isMatchingService(service, discoveredService))) {
+        services.remove(service);
+      }
+    }
+
+    return services.toList();
```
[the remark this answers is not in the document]
spkersten: It is the same instance as `_services` and I don't want to give clients a mutable private field. (A `ListView` could be a way to avoid the copy but that seems a bit uncommon and the list isn't big)

packages/reactive_ble_mobile/ios/Classes/BleData extras/CharacteristicInstance.swift
```diff
@@ -0,0 +1,99 @@
...
+        }
+        
+        guard
+            let characteristicIndex = service.characteristics?.filter({ c in c.uuid == characteristic.uuid }).index(of: characteristic)
+        else {
+            throw Failure.characteristicNotFound
+        }
+        
+        guard
+            let serviceIndex = peripheral.services?.filter({ s in s.uuid == service.uuid }).index(of: service)
```
gabrielgarciagava, line 34: If I understand correctly, and also based on our discussions, on iOS we indentify the services by its order in the list of services for some specific characteristic uuid.

Should it get some comment to explain that for the serviceIndex, or is this information useless in the point of view of the public interface?

Also, confusing that you need to filter the `peripheral.services`, since the peripheral comes from `service.peripheral` in the first place

packages/reactive_ble_mobile/ios/Classes/ReactiveBle/Central.swift
```diff
@@ -82,17 +84,27 @@ final class Central {
...
+                
                 central.characteristicNotifyRegistry.updateTask(
-                    key: QualifiedCharacteristic(characteristic),
+                    key: q,
                     action: { $0.complete(error: error) }
                 )
             },
             onCharacteristicValueUpdate: papply(weak: self) { central, characteristic, error in
-                onCharacteristicValueUpdate(central, QualifiedCharacteristic(characteristic), characteristic.value, error)
+                onCharacteristicValueUpdate(central, try! CharacteristicInstance(characteristic), characteristic.value, error)
```
gabrielgarciagava, line 98: Not clear to me why here you force try building the CharacteristicInstance and not on the other two methods.

packages/reactive_ble_mobile/ios/Classes/BleData extras/CharacteristicInstance.swift
```diff
@@ -0,0 +1,99 @@
...
+        else {
+            throw Failure.serviceNotFound
+        }
+        
+        self.init(
+            id: characteristic.uuid,
+            instanceID: "\(characteristicIndex)",
```
gabrielgarciagava, line 41: It looks like here you go from integer to InstanceID, and then back to integer where you need to use it. Is it a good reason for it to ever be a InstanceID? Just asking that because in the place where you convert it back to Int it is not completly obvious why you can do that.

packages/reactive_ble_mobile/ios/Classes/ReactiveBle/Tasks/ServicesWithCharacteristicsDiscovery/ServicesWithCharacteristicsDiscoveryTaskController.swift
```diff
@@ -101,9 +101,9 @@ struct ServicesWithCharacteristicsDiscoveryTaskController: PeripheralTaskControl
             items = servicesWithCharacteristicsToDiscover
                 .compactMap { item in
                     knownServices?
-                        .first(where: { $0.uuid == item.key })
-                        .flatMap { service in (service, item.value) }
-                }
+                        .filter({ $0.uuid == item.key })
+                        .map { service in (service, item.value) }
+                }.flatMap({ x in x })
```
gabrielgarciagava, line 106: nit:
You can probably write `.flatMap{ $0 }` or `.joined()`

packages/reactive_ble_platform_interface/lib/src/model/characteristic_instance.dart
```diff
@@ -0,0 +1,52 @@
+import 'package:meta/meta.dart';
+
+import 'uuid.dart';
+
+@immutable
+class CharacteristicInstance {
+  /// Unique uuid of the specific characteristic
+  final Uuid characteristicId;
+
+  /// When empty, the first characteristic with [characteristicId] is used
```
gabrielgarciagava, line 10: When you say empty you are talking about empty string? Sorry, got a little bit confused with it.

gabrielgarciagava: And if so, why decided to go with empty string instead of nullability?

gabrielgarciagava (approved):

packages/flutter_reactive_ble/lib/src/reactive_ble.dart
```diff
@@ -154,24 +152,28 @@ class FlutterReactiveBle {
   /// Be aware that a read request could be satisfied by a notification delivered
   /// for the same characteristic via [characteristicValueStream] before the actual
   /// read response arrives (due to the design of iOS BLE API).
-  Future<List<int>> readCharacteristic(
-      QualifiedCharacteristic characteristic) async {
+  ///
+  /// This method assumes there is a single characteristic with the ids specified in [characteristic]. If there are
+  /// multiple characteristics with the same id on a device, use [resolve] to find them all. Or use
+  /// [getDiscoveredServices] to select the [Service]s and [Characteristic]s you're interested in.
+  Future<List<int>> readCharacteristic(QualifiedCharacteristic characteristic) async {
```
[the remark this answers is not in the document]
spkersten: I see no compelling reason to deprecate them. It would force the majority of users who deal with devies without multiple services/chars of the same type to refactor their code.

packages/flutter_reactive_ble/lib/src/reactive_ble.dart
```diff
@@ -290,54 +294,258 @@ class FlutterReactiveBle {
...
+            lib: this,
+            isReadable: discoveredCharacteristic.isReadable,
+            isWritableWithoutResponse: discoveredCharacteristic.isWritableWithoutResponse,
+            isWritableWithResponse: discoveredCharacteristic.isWritableWithResponse,
+            isNotifiable: discoveredCharacteristic.isNotifiable,
+            isIndicatable: discoveredCharacteristic.isIndicatable,
+          ));
+        }
+      }
+      for (final char in service._characteristics.toList()) {
```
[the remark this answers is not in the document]
spkersten: `service._characteristics` is final, but I've changed it to use `removeWhere` instead of using a for-loop.

packages/reactive_ble_platform_interface/lib/src/model/characteristic_instance.dart
spkersten: That comment isn't relevant/correct anymore. I'll replace it with an explanation of what instance id is.

packages/reactive_ble_mobile/ios/Classes/BleData extras/CharacteristicInstance.swift
spkersten: I made it a string because I don't know what would be a convenient format on other platforms. Within both the Android and iOS implementation it is known to be (convertible to) an int.

packages/reactive_ble_mobile/ios/Classes/BleData extras/CharacteristicInstance.swift
spkersten: It's an implementation detail of the Swift code that the index is used.

I'm not sure what is confusing about filtering `peripheral.services`. It contains services of all kinds and I'm interested in the index within the services of the same kind (hoping that number is stable).

gabrielgarciagava (approved):

Taym95 (approved):

castehlin:
Hi,

Sorry if this is the wrong place since it's closed, but I haven't found a place that covers this latest change and if it addresses characteristics with identical service and char uuids, but different properties (write/read/notify etc). The documentation for this package doesn't seem to cover this on pub.dev.

Provided I've found the unique characteristic using the isWritable/isNotifyable fields or the  instance_id, how can I get something like writeCharacteristic() or subscribeCharacteristic() to take in these unique identifiers, rather than just what the QualifiedCharacteristic object allows (only charUuid, serviceUuid, deviceId?)

Is this supported in this package yet? 

SO post that covers this in a bit more detail:
https://stackoverflow.com/questions/77052224/how-to-write-notify-to-characteristic-with-identical-service-and-char-uuids-in-f

spkersten:
@castehlin The entry point for dealing with services or characteristics that don't have a unique id is `getDiscoveredServices`

PR #778, opened 2023-08-05
Fix typos

wcoder:
While using the package from sources, some typos were found and fixed to make the package more mature, I'm giving back those fixes.

Thanks for the package.


Taym95:
Thank you!

PR #800, opened 2023-09-27
Implement all platforms

rohitsangwan01:
Implemented MacOS support using same Swift code, using [sharedDarwinSource](https://docs.flutter.dev/packages-and-plugins/developing-packages#shared-ios-and-macos-implementations)

We can use `darwin` folder instead of ios or macos and keep the same source code for both Apple platforms, so this way i guess maintenance efforts would be same


rohitsangwan01:
Opps, wrong one

PR #801, opened 2023-09-27
Implement MacOS support using sharedDarwinSource

rohitsangwan01:
Implemented MacOS support using same Swift code, using [sharedDarwinSource](https://docs.flutter.dev/packages-and-plugins/developing-packages#shared-ios-and-macos-implementations)

We can use darwin folder instead of ios or macos and keep the same source code for both Apple platforms, so this way i guess maintenance efforts would be same


Taym95:
@rohitsangwan01 did you test all functionally in macos?

rohitsangwan01:
@Taym95 i tried basic tasks like connect, disconnect and some services related tasks, but not in detail, and because macos apis are exactly similar to IOS, so both should behave similar

Taym95:
> @Taym95 i tried basic tasks like connect, disconnect and some services related tasks, but not in detail, and because macos apis are exactly similar to IOS, so both should behave similar in functionality

I tried this in the past, and not all API worked, please test other functionalities

rohitsangwan01:
@Taym95 ok sure, will test all functionalities

PR #810, opened 2023-10-24
Fix typo in log message disconnect in example

mdxs:
Fixed typo in log message disconnect in example: changed "disconnecting to device:" into the more proper "disconnecting from device:" (similar to log message in exception part).


mdxs:
@Taym95 Are there specific PR contribution rules that I didn't follow?

Issue #835, opened 2024-01-26
'readRssi' Method Not Defined for 'ReactiveBlePlatform'


DeToxer:
**Describe the bug**
After upgrading the `flutter_reactive_ble` library from version `5.2.0` to `5.3.0,` the app fails to compile on both Android and iOS platforms. The compiler error indicates that the method `readRssi` isn't defined for the class `ReactiveBlePlatform`. This issue has been tested and confirmed on two devices: an iPhone simulator (iOS version 17.0, iPhone 15 Pro Max) and an Android emulator (Pixel 7 Pro API 33). The problem persists despite attempts to clean the build environment using `flutter clean `and `flutter pub cache clean --force`. Notably, this issue does not arise in a newly created Flutter project.

**Smartphone / tablet**
Device: iPhone 15 Pro Max/Pixel 7 Pro
OS: iOS 17.0/Android 12 (SDK 33)
Package version: `flutter_reactive_ble 5.3.0`

**Additional context**
The issue appears to be specifically associated with version 5.3.0 of the `flutter_reactive_ble library`. Reverting to version 5.2.0 or initiating a new Flutter project does not result in this compile-time error, suggesting that the problem is isolated to this particular version update.

Android stack trace:
```
- 'ReactiveBlePlatform' is from 'package:reactive_ble_platform_interface/src/reactive_ble_platform_interface.dart' ('../../../../../.pub-cache/hosted/pub.dev/reactive_ble_platform_interface-5.2.0/lib/src/reactive_ble_platform_interface.dart').
Try correcting the name to the name of an existing method, or defining a method named 'readRssi'.
      _blePlatform.readRssi(deviceId);
                   ^^^^^^^^
Target kernel_snapshot failed: Exception


FAILURE: Build failed with an exception.
```

iOS stack tace:
```
Error (Xcode): ../../../../../.pub-cache/hosted/pub.dev/flutter_reactive_ble-5.3.0/lib/src/reactive_ble.dart:405:20: Error: The method 'readRssi' isn't defined for the class 'ReactiveBlePlatform'.

Could not build the application for the simulator.
Error launching application on iPhone 15 Pro Max.
```

**Flutter doctor -v output log**
```
[✓] Flutter (Channel stable, 3.16.9, on macOS 14.1.1 23B81 darwin-arm64, locale nb-NO)
    • Flutter version 3.16.9 on channel stable at /Users/detoxer/dev/sdk/flutter
[...]

PR #837, opened 2024-01-29
Use proper platform_interface dependency version

vbuberen:
Update required version of the plugin to fix 5.3.0 release.

Fixes https://github.com/PhilipsHue/flutter_reactive_ble/issues/835

This is annoying issue and we had it quite often in Plus Plugins earlier before.

### DIFF
--- a/packages/flutter_reactive_ble/pubspec.yaml
+++ b/packages/flutter_reactive_ble/pubspec.yaml
@@ -21,8 +21,8 @@ dependencies:
     sdk: flutter
   functional_data: ^1.0.0
   meta: ^1.3.0
-  reactive_ble_mobile: ^5.2.0
-  reactive_ble_platform_interface: ^5.2.0
+  reactive_ble_mobile: ^5.3.0
+  reactive_ble_platform_interface: ^5.3.0
 
 dependency_overrides:
   reactive_ble_mobile:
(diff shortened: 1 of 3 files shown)


packages/flutter_reactive_ble/pubspec.yaml
```diff
@@ -21,8 +21,8 @@ dependencies:
     sdk: flutter
   functional_data: ^1.0.0
   meta: ^1.3.0
-  reactive_ble_mobile: ^5.2.0
-  reactive_ble_platform_interface: ^5.2.0
+  reactive_ble_mobile: ^5.3.0
+  reactive_ble_platform_interface: ^5.3.0
```
Taym95, line 25: please also update example/pubspec.yaml app

vbuberen: Done, but it doesn't make much sense to delay merging and releasing only because of this, considering that example app has `dependency_overrides` pointing to local version of packages 😄

Taym95: Let's do it right, also can you please run `./bin/quality_checks.sh`

Taym95: `pubspec.lock` did not change

Taym95 (approved):

packages/flutter_reactive_ble/pubspec.yaml
Taym95: Indeed , Thanks lot!

PR #848, opened 2024-02-28
Update PluginController.swift so it can be debugged

horner:
the assert blocks debugging in Flutter with VSCode when breakpoints were enabled.  Video to come

### DIFF
--- a/packages/reactive_ble_mobile/ios/Classes/Plugin/PluginController.swift
+++ b/packages/reactive_ble_mobile/ios/Classes/Plugin/PluginController.swift
@@ -38,7 +38,7 @@ final class PluginController {
             },
             onDiscovery: papply(weak: self) { context, _, peripheral, advertisementData, rssi in
                 guard let sink = context.scan?.sink
-                else { assert(false); return }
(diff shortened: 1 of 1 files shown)


packages/reactive_ble_mobile/ios/Classes/Plugin/PluginController.swift
```diff
@@ -38,7 +38,7 @@ final class PluginController {
             },
             onDiscovery: papply(weak: self) { context, _, peripheral, advertisementData, rssi in
                 guard let sink = context.scan?.sink
-                else { assert(false); return }
```
Taym95, line 41: @werediver can you give it quick look?

werediver: The assert is there to indicate a weird condition early. It only fires in debug builds.

If one is down to _debugging_ the library code, they are free to use a dependency override with a locally modified copy.

I'd prefer to keep the code as is (reject this p/r), unless a strong and sound rationale is given supporting this change.

Issue #872, opened 2024-06-14
connectToAdvertisingDevice while already scanning


maBarabas:
**Describe the bug**

When connecting to a device while a scan is in progress, TimeoutException is thrown.

It looks like the intention was that the existing scan results can be reused for the connection and the connection should work. However, this only happens if the existing scan ends within the timeout [here](https://github.com/PhilipsHue/flutter_reactive_ble/blame/96e083ac9edd2b6cb070f6534bb2309057257aa0/packages/flutter_reactive_ble/lib/src/device_connector.dart#L168). If the scan ends early, then connection is attempted before prescanDuration has elapsed, so it seems like it doesn't work as intended. If the scan doesn't end within the timeout, then the connection is never attempted even if the device was scanned by the existing scan session.

**To Reproduce**
1. Scan for devices
1. Connect to device using `connectToAdvertisingDevice()` while keeping the scan going
1. Connection fails with a prescanDuration + 1s timeout, and no stack trace

**Expected behavior**
Existing scan results are used to connect to the advertising device, instead of waiting for the existing scan to end in prescanDuration + 1s, then failing if it doesn't end.

- [ ] ~~I tried doing the same with a general BLE scanner application (e.g. [nRF Connect](https://duckduckgo.com/?q=nRF+connect)) and it exhibits the expected behavior as described above~~ This issue is related to the library code

**Smartphone / tablet**
 - Device: any, issue is in Dart code
 - OS: any, issue is in Dart code
 - Package version: any, same code exists in the initial commit

**Peripheral device**
doesn't matter

**Additional context**

- The other branch running when there is no existing scan in progress has an early check to connect for devices scanned by a recenlty completed scan sessions: https://github.com/PhilipsHue/flutter_reactive_ble/blame/96e083ac9edd2b6cb070f6534bb2309057257aa0/packages/flutter_reactive_ble/lib/src/device_connector.dart#L198. This is missing from the branch that runs when a scan is already in progress.

PR #875, opened 2024-06-14
`connectToAdvertisingDevice`: Check for recently discovered device when BLE scan already running

maBarabas:
Prevents error reported in #872 as long as the device was scanned within the last `_scanRegistryCacheValidityPeriod` (25 seconds) from where `connectToAdvertisingDevice` is called. In practice, this is as good as always for usecases where a device is selected from a scan list and immediately connected.

### DIFF
--- a/packages/flutter_reactive_ble/lib/src/device_connector.dart
+++ b/packages/flutter_reactive_ble/lib/src/device_connector.dart
@@ -85,7 +85,17 @@ class DeviceConnectorImpl implements DeviceConnector {
     Map<Uuid, List<Uuid>>? servicesWithCharacteristicsToDiscover,
     Duration? connectionTimeout,
   }) {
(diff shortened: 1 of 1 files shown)


packages/flutter_reactive_ble/lib/src/device_connector.dart
```diff
@@ -85,7 +85,17 @@ class DeviceConnectorImpl implements DeviceConnector {
     Map<Uuid, List<Uuid>>? servicesWithCharacteristicsToDiscover,
     Duration? connectionTimeout,
   }) {
-    if (_deviceScanner.currentScan != null) {
+    if (_deviceIsDiscoveredRecently(
```
Taym95, line 88: why don't you just stop the scan and connect with `connectToAdvertisingDevice ` ?

Issue #934, opened 2026-05-18
Kotlin Gradle Plugin (KGP) warning on Android build after upgrading to Flutter 3.44


TheJanzap:
After migrating to Flutter 3.44 with Dart 3.12 and applying the Kotilin Migration Guide for App developers of the ["Migrating Flutter Android projects to built-in Kotlin" breaking change](https://docs.flutter.dev/release/breaking-changes/migrate-to-built-in-kotlin), the following warning appears when buidling for Android:

```
WARNING: Your app uses the following plugins that apply Kotlin Gradle Plugin (KGP): reactive_ble_mobile
Future versions of Flutter will fail to build if your app uses plugins that apply KGP.

Please check the changelogs of these plugins and upgrade to a version that supports Built-in Kotlin.
If no such version exists, report the issue to the plugin. If necessary, here is a guide on filing
an issue against a plugin: https://docs.flutter.dev/release/breaking-changes/migrate-to-built-in-kotlin/for-app-developers#report-incompatible-kotlin-gradle-plugin-usage-to-plugin-authors

If you are a plugin author, please migrate your plugin to Built-in Kotlin using this guide: https://docs.flutter.dev/release/breaking-changes/migrate-to-built-in-kotlin/for-plugin-authors
```

Flutter advises that I pass this message onto you: 

> Starting with Android Gradle Plugin (AGP) 9.0, support for applying the Kotlin Gradle Plugin (KGP) has been removed. Because this plugin applies KGP, it causes a compilation error that prevents my app from building. Here is an example of the error [Issue #181383](https://github.com/flutter/flutter/issues/181383).
Flutter has temporarily added support to allow KGP while apps and plugins migrate to AGP 9.0+, but this support will be removed in a future version of Flutter.
Please migrate this plugin to use built-in Kotlin to ensure your plugin users can successfully build their apps in future versions of Flutter.
Here is the Flutter [migration guide for plugin authors](https://docs.flutter.dev/release/breaking-changes/migrate-to-built-in-kotlin/for-plugin-authors).

Would it be possible to apply the steps in the migration guide listed above? Thank you!

PR #937, opened 2026-08-04
fix(android): update compileSdk to 35 and Java compatibility to 17

syedabdulbasitali1:
### Summary
Fixes Android compilation errors on modern Flutter versions (Flutter 3.22+) and modern JDK environments.

### Changes
- Updated `compileSdk` to `35` in `packages/reactive_ble_mobile/android/build.gradle` to fix `checkDebugAarMetadata` errors with modern AndroidX dependencies (`core:1.13.1`, `fragment:1.7.1`).
- Set `sourceCompatibility`, `targetCompatibility`, and `jvmTarget` to Java 17 to fix obsolete Java 8 build warnings.

### Testing
- Tested in an isolated Flutter test app with `flutter run`.
- Confirmed the app builds and installs successfully on an Android emulator without build errors or Java warnings.

PR #935, opened 2026-05-23
Update Android compileSdkVersion from 33 to 34

juanchavez4096:
Fixes issue https://github.com/PhilipsHue/flutter_reactive_ble/issues/926 https://github.com/PhilipsHue/flutter_reactive_ble/issues/911
this change adds modern android compatibility, fixing androidx issue

PR #938, opened 2026-08-17
Support built-in Kotlin, AGP 9.0+ and Android SDK 37

tseybold-beurer:
Fix issue #934

- Migrate to built-in Kotlin
- Upgrade the following versions:
    - Android SDK 37
    - Kotlin 2.4.10
    - AGP 9.1.0
    - Gradle 9.3.1
    - Java 17
- Set min Flutter version to 3.44.0

Tested with integrated example and a production app.

Versions match Android dependency matrix defined in https://flutter.dev/blog/whats-new-in-flutter-3-47

### DIFF
--- a/packages/reactive_ble_mobile/pubspec.lock
+++ b/packages/reactive_ble_mobile/pubspec.lock
@@ -327,10 +327,10 @@ packages:
     dependency: transitive
     description:
       name: matcher
-      sha256: dc0b7dc7651697ea4ff3e69ef44b0407ea32c487a39fff6a4004fa585e901861
+      sha256: "31bd099b47c10cd1aeb55146a2d46ce0277630ecef3f7dae54ad7873f36696cd"
       url: "https://pub.dev"
     source: hosted
-    version: "0.12.19"
+    version: "0.12.20"
   material_color_utilities:
     dependency: transitive
     description:
@@ -343,10 +343,10 @@ packages:
     dependency: transitive
     description:
       name: meta
-      sha256: "23f08335362185a5ea2ad3a4e597f1375e78bce8a040df5c600c8d3552ef2394"
+      sha256: "307249ce4ff29d58a18e97f6345f539382eb9c9c29ecda628900f31de0443dd9"
       url: "https://pub.dev"
     source: hosted
-    version: "1.17.0"
+    version: "1.19.0"
   mime:
     dependency: transitive
     description:
@@ -507,10 +507,10 @@ packages:
     dependency: transitive
     description:
       name: test_api
-      sha256: "8161c84903fd860b26bfdefb7963b3f0b68fee7adea0f59ef805ecca346f0c7a"
+      sha256: "2a122cbe059f8b610d3a5415f42e255b6c17b1f21eee1d960f31080237fb4f11"
       url: "https://pub.dev"
     source: hosted
-    version: "0.7.10"
+    version: "0.7.12"
   timing:
     dependency: transitive
     description:
@@ -531,10 +531,10 @@ packages:
     dependency: transitive
     description:
       name: vector_math
-      sha256: d530bd74fea330e6e364cda7a85019c434070188383e1cd8d9777ee586914c5b
+      sha256: f36f9f3be64c6198714492bb455c11056e33e2f85d9a0b676a48301e44fdcf47
       url: "https://pub.dev"
     source: hosted
-    version: "2.2.0"
+    version: "2.4.2"
   vm_service:
     dependency: transitive
     description:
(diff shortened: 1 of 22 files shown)


tseybold-beurer:
Update everything to current versions, which makes #935 and #937 obsolete

tseybold-beurer:
@werediver @PieterAelse @Taym95 could you please review this change?

This is a necessary change so this lib can still be used with current Flutter versions.

Sorry for pinging you directly, but there was no update/response in this repo since 3 month.

werediver:
The library works as is with Flutter 3.44.9, just for information. The changes appear to be mostly Android-specific, so I'd prefer if @Taym95 and/or @PieterAelse look at those.

tseybold-beurer:
> The library works as is with Flutter 3.44.9, just for information. The changes appear to be mostly Android-specific, so I'd prefer if @Taym95 and/or @PieterAelse look at those.

Thanks for the fast answer :)

You are completely right, the lib works with current Flutter versions.
But with every build, Flutter throws red warnings that the project should be updated to built-in Kotlin (#934) and to Gradle 9+. This is not possible without an update of this lib.

And yes, its an Android-specific change.

PieterAelse:
@tseybold-beurer Thanks for this PR! I tried to run the required checks but one fails on the old Flutter version. Can you update `./github/workflows/main.yml` to use `flutter-version: '3.44.0'` ? Then the build can be run 😄

PieterAelse:
@tseybold-beurer there appears to be some build issue(s), could you have a look? https://github.com/PhilipsHue/flutter_reactive_ble/actions/runs/32340394130/job/96704859126?pr=938

tseybold-beurer:
> @tseybold-beurer there appears to be some build issue(s), could you have a look? https://github.com/PhilipsHue/flutter_reactive_ble/actions/runs/32340394130/job/96704859126?pr=938

@PieterAelse  Tested with the pipelines in my local repo. Should now work fine.

PieterAelse:
@tseybold-beurer Thank you for fixing lint/test. I'm not sure about bumping the min-Flutter version to 3.47 though, why is that needed? If everything in this MR works find on 3.44, lets leave it at that. (CC @werediver  / @Taym95 )

tseybold-beurer:
@PieterAelse Sure, I set it to 3.44 again :)

We then just have to set `android.builtInKotlin=false` to true, which requires 3.47, in a future release.

packages/reactive_ble_mobile/pubspec.lock
```diff
@@ -577,4 +577,4 @@ packages:
     version: "3.1.2"
 sdks:
   dart: ">=3.11.0 <4.0.0"
-  flutter: ">=3.41.0"
+  flutter: ">=3.47.0"
```
PieterAelse: I think this should go back to `3.44.0`, right?

PieterAelse:
> @PieterAelse Sure, I set it to 3.44 again :)
> 
> We then just have to set `android.builtInKotlin=false` to true, which requires 3.47, in a future release.

Yeah let's tackle that once we get there! For now people on Flutter < 3.47 should still be happy using the library.

I left one small final comment and then I think I can merge 😄

alxwo:
I tested the current PR (`a424025`) with:

- Flutter 3.47.1 / Dart 3.13.1
- AGP 9.1.1
- Gradle 9.3.1
- compileSdk/targetSdk 37
- android.builtInKotlin=true

The unchanged PR fails in `:reactive_ble_mobile:compileDebugKotlin` because it pulls in `kotlin-stdlib 2.4.10`, while the compiler used by AGP built-in Kotlin expects older metadata.

The build succeeds after removing these plugin-local declarations from `reactive_ble_mobile/android/build.gradle`:

- `ext.kotlin_version = '2.4.10'`
- `classpath "org.jetbrains.kotlin:kotlin-gradle-plugin:$kotlin_version"`
- `implementation "org.jetbrains.kotlin:kotlin-stdlib-jdk7:$kotlin_version"`

I kept the existing `kotlin { compilerOptions { ... JVM_17 } }` block unchanged.

After that the debug APK builds and installs successfully with `android.builtInKotlin=true`.

Maybe it would make sense to let AGP built-in Kotlin provide the Kotlin compiler/runtime instead of declaring KGP and stdlib explicitly in the plugin.

tseybold-beurer:
@PieterAelse I updated the pubspec.lock 👍 

@alxwo In the PR you tested, builtInKotlin should be set to false.

Setting this to true, and therefore completely removing the Kotlin version, requires setting Flutter min version to 3.47. I did this in an earlier change, but @PieterAelse recommended to better keep minimum at 3.44

PieterAelse (approved):

GregoryFournier:
Hi, when will these changes get added to a release?

PieterAelse:
@GregoryFournier Thanks for the notification! Seems I forgot about it (or low on my TODO-list). I just published 🎉 

See https://pub.dev/packages/flutter_reactive_ble/changelog#560 

CC @tseybold-beurer / @alexwastl

